PR #3, opened 2019-01-28
Switch clair-local-scan image from versioned to latest

eversC:


eversC:
I've tested this locally, seems to work fine.

Issue #11, opened 2019-02-20
New workspace idempotency


olib963:
I think it would be good to check the `terraform workspace list` before creating a new workspace since terraform is not idempotent itself.

PR #12, opened 2019-02-21
Check workspace before create

olib963:
Addressing https://github.com/ovotech/circleci-orbs/issues/11. @danielflookovo I can EOF it if you think that's nicer.

### DIFF
--- a/terraform/orb.yml
+++ b/terraform/orb.yml
@@ -5,7 +5,7 @@ executors:
   default:
     description: "Terraform executor"
     docker:
-    - image: ovotech/terraform:latest
+      - image: ovotech/terraform:latest
 
 commands:
   plan:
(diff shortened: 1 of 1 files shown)


terraform/orb.yml
```diff
@@ -211,7 +211,11 @@ commands:
           name: "terraform workspace new"
           command: |
             include init.sh
-            exec terraform workspace new "$workspace" "$module_path"
+            exec if [[ -z "$(terraform workspace list $module_path | grep $workspace)" ]]; then \
+              terraform workspace new "$workspace" "$module_path"; \
```
danielflookovo: Can you `exec terraform`, not the if. Doesn't matter much here, but for consistency

olib963: Done :D

PR #22, opened 2019-05-28
Add parameter for controlling the exit code

skjolber:
We're running clair-scanning on a set of images and want to do some post-processing (i.e. parsing and communicating the results). So it would be helpful if failing the job is optional.

### DIFF
--- a/clair-scanner/orb.yml
+++ b/clair-scanner/orb.yml
@@ -30,7 +30,7 @@ commands:
       fail_on_discovered_vulnerabilities:
         type: "boolean"
         description: "Fail command when vulnerabilities at severity equal to or above the threshold are discovered"
-        default: "true"
+        default: true
     steps:
       - checkout
       - setup_remote_docker:
--- a/clair-scanner/orb.yml
+++ b/clair-scanner/orb.yml
@@ -25,8 +25,12 @@ commands:
         default: ""
       severity_threshold:
         type: "string"
-        description: "The threshold above which discovered vulnerabilities will fail the build. May be 'Defcon1', 'Critical', 'High', 'Medium', 'Low', 'Negligible' or 'Unknown'"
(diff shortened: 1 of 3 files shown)


clair-scanner/scan.sh
```diff
@@ -48,4 +48,8 @@ else
     scan "<< parameters.image >>"
 fi
 
-exit $EXIT_STATUS
+if [ "<< fail_on_discovered_vulnerabilities >>" = "false"]
```
skjolber: Not completely sure how the boolean parameter type plays here.

clair-scanner/orb.yml
```diff
@@ -25,8 +25,12 @@ commands:
         default: ""
       severity_threshold:
         type: "string"
-        description: "The threshold above which discovered vulnerabilities will fail the build. May be 'Defcon1', 'Critical', 'High', 'Medium', 'Low', 'Negligible' or 'Unknown'"
+        description: "The threshold (equal and above) at which discovered vulnerabilities are reported. May be 'Defcon1', 'Critical', 'High', 'Medium', 'Low', 'Negligible' or 'Unknown'"
         default: "High"
+      fail_on_discovered_vulnerabilities:
+        type: "boolean"
+        description: "Fail command when vulnerabilities at severity equal to or above the threshold are discovered"
+        default: "true"
```
danielflookovo: This needs to be unquoted to be a boolean

```suggestion
        default: true
```

clair-scanner/scan.sh
danielflookovo: CircleCI needs `parameters.` before the parameter name. 
Bash needs a space before the `]`, and `==` for the comparison.

```suggestion
if [ "<< parameters.fail_on_discovered_vulnerabilities >>" == "false" ]; then
```

skjolber:
@danielflookovo thanks for QA / release! :1st_place_medal:

PR #23, opened 2019-05-29
Move clair reports into workspace directory

skjolber:
Persisting to workspace does not work out of the box without a mv / cp command; error message says that path must be relative to workspace. So move it to workspace and rather store artifacts from there.

### DIFF
--- a/clair-scanner/orb.yml
+++ b/clair-scanner/orb.yml
@@ -40,4 +40,4 @@ commands:
           command: |
             include scan.sh
       - store_artifacts:
-          path: /clair-reports
+          path: clair-reports
--- a/clair-scanner/scan.sh
+++ b/clair-scanner/scan.sh
@@ -48,6 +48,8 @@ else
     scan "<< parameters.image >>"
 fi
 
+mv $REPORT_DIR .
+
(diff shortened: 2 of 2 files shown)


eversC:
Hi @skjolber 

I'm struggling to replicate the issue here.

Just as a test, I've specified the following job that scans images:

```
  scan_images:
    executor: clair_scanner/default
    steps:
    - clair_scanner/scan:
        image: debian:buster
    - persist_to_workspace:
        root: /
        paths:
          - clair-reports
```

and the following job (later in the workflow) to check whether the workspace can be attached properly:

```
  ls_workspace:
    executor: clair_scanner/default
    steps:
    - attach_workspace:
        at: /tmp/workspace
    - run: |
        ls -l /tmp/workspace/clair-reports
```

I got the following output in CircleCI from the 2nd job:

```
#!/bin/bash -eo pipefail
ls -l /tmp/workspace/clair-reports

total 28
-rw-r--r-- 1 root root 24885 Oct  3 22:42 debian_buster.json

CircleCI received exit code 0
```

Am I missing something?

skjolber:
Ok, guess I missed that root parameter. Thanks

PR #25, opened 2019-05-31
Add support for the -target arg

mikebrooks-ovo:


terraform/orb.yml
```diff
@@ -215,7 +219,7 @@ commands:
               exec terraform workspace new "$workspace" "$module_path"
             else
               echo "$workspace already exists"
-            fi 
+            fi
```
danielflookovo: 👀

mikebrooks-ovo: "format on save" :D

danielflookovo:
👍 Can you add it for the plan and check jobs/commands, and update the readme

terraform/apply.sh
```diff
@@ -13,6 +13,12 @@ EOF
 
 set +e
 
+if [ "<< parameters.auto_approve >>" = "true" && -n "<< parameters.target >>" ]; then
```
danielflookovo: I don't think thats valid bash
```suggestion
if [[ "<< parameters.auto_approve >>" = "true" && -n "<< parameters.target >>" ]]; then
```

PR #28, opened 2019-06-13
Add AWS rotate keys orb.

mwz:
This PR adds a new ORB for rotating AWS keys.

### DIFF
--- a/aws-rotate-keys/README.md
+++ b/aws-rotate-keys/README.md
@@ -1,20 +1,16 @@
 # AWS CircleCI key rotate orb
 
-This orb can be used to rotate AWS access keys and update corresponding CircleCI environment variables.
-A common use case would be to set up a scheduled CircleCI job which rotates access keys for the used account.
+This orb can be used to rotate AWS access keys and update corresponding CircleCI environment variables. A common use case would be to set up a scheduled CircleCI job which rotates access keys for the used user.
 
 ## Executors
-This orb defines a small 'default' executor for running the aws commands.
-Any container that has aws cli and curl can run the this orb.
+This orb defines a small 'default' executor for running the aws commands. Any container that has `aws` cli, `curl` and `jq` can be used to run this orb.
 
 ## Commands
 ### rotate
-This is the only command available in this orb. It rotates the AWS access keys for the account specified as the `aws-username` parameter. In addition of rotating the keys, this command also updates the corresponding environment variables. In order to run this command, you need to make sure that the aws cli client is already authenticated.
+This is the only command available in this orb. It rotates the AWS access keys for the user specified as the `aws-username` parameter. In addition of rotating the keys, this command also updates the corresponding environment variables. In order to run this command, you need to make sure that the aws cli client is already authenticated.
 
 **Parameters**
 - `aws-username` - user name of the AWS account you want to rotate keys for
-- `github-org` - name of the Github organisation the repository belongs to
-- `github-repo` - name of the Github repository
 - `circleci-token` - CircleCI API token used to update environment variables
 - `aws-access-key-id-var` - name of the CircleCI environment variable which holds a value of the aws access key id, e.g. `AWS_ACCESS_KEY_ID`
 - `aws-secret-access-key-var` - name of the CircleCI environment variable which holds a value of the aws secret access key, e.g. `AWS_SECRET_ACCESS_KEY`. 
@@ -34,11 +30,10 @@ orbs:
 
 jobs:
   rotate-aws-keys:
+    executor: rotate-aws-keys/default
     steps:
       - rotate-aws-keys/rotate:
           aws-username: circleci-user
-          github-org: github-org
-          github-repo: github-repo-name
           circleci-token: $CIRCLECI_TOKEN
           aws-access-key-id-var: AWS_ACCESS_KEY_ID
           aws-secret-access-key-var: AWS_SECRET_ACCESS_KEY
@@ -64,6 +59,7 @@ orbs:
 
 jobs:
   rotate-aws-keys:
+    executor: rotate-aws-keys/default
     steps:
       - run:
           command: |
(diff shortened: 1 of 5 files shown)


mwz:
cc @ovotech/green, @ovotech/production-eng

aws-rotate-keys/README.md
```diff
@@ -0,0 +1,90 @@
...
+The following example uses the following CircleCI env variables `PROD_AWS_ACCESS_KEY_ID` and `PROD_AWS_SECRET_ACCESS_KEY` which are not automatically picked up by aws cli and therefore you need to set up the credentials manually.
+
+```yaml
+version: 2.1
+
+orbs:
+  rotate-aws-keys: ovotech/aws-rotate-keys@1
+
+jobs:
+  rotate-aws-keys:
```
danielflookovo: Does this work? I don't see any executor specified

aws-rotate-keys/README.md
```diff
@@ -0,0 +1,90 @@
...
+This orb defines a small 'default' executor for running the aws commands.
+Any container that has aws cli and curl can run the this orb.
+
+## Commands
+### rotate
+This is the only command available in this orb. It rotates the AWS access keys for the account specified as the `aws-username` parameter. In addition of rotating the keys, this command also updates the corresponding environment variables. In order to run this command, you need to make sure that the aws cli client is already authenticated.
+
+**Parameters**
+- `aws-username` - user name of the AWS account you want to rotate keys for
+- `github-org` - name of the Github organisation the repository belongs to
```
danielflookovo: I assume this is so you can rotate the keys in a different repo to the one using the orb.
CircleCI exposes the current org/repo as environment variables, which would be good defaults.
I doubt anyone would need to change the github-org, for example.

aws-rotate-keys/README.md
```diff
@@ -0,0 +1,90 @@
...
+- `github-repo` - name of the Github repository
+- `circleci-token` - CircleCI API token used to update environment variables
+- `aws-access-key-id-var` - name of the CircleCI environment variable which holds a value of the aws access key id, e.g. `AWS_ACCESS_KEY_ID`
```
danielflookovo: Can we also use the standard aws var names as defaults

aws-rotate-keys/README.md
```diff
@@ -0,0 +1,90 @@
...
+This orb can be used to rotate AWS access keys and update corresponding CircleCI environment variables.
+A common use case would be to set up a scheduled CircleCI job which rotates access keys for the used account.
+
+## Executors
+This orb defines a small 'default' executor for running the aws commands.
+Any container that has aws cli and curl can run the this orb.
+
+## Commands
+### rotate
+This is the only command available in this orb. It rotates the AWS access keys for the account specified as the `aws-username` parameter. In addition of rotating the keys, this command also updates the corresponding environment variables. In order to run this command, you need to make sure that the aws cli client is already authenticated.
```
danielflookovo: Can you change the word account to user. The AWS account is different, and implied by the access key. Sorry to be pedantic, but I had to read this a few times to understand.

```suggestion
This is the only command available in this orb. It rotates the AWS access keys for the user specified as the `aws-username` parameter. In addition of rotating the keys, this command also updates the corresponding environment variables. In order to run this command, you need to make sure that the aws cli client is already authenticated.
```

aws-rotate-keys/README.md
```diff
@@ -0,0 +1,90 @@
...
+
+**Parameters**
+- `aws-username` - user name of the AWS account you want to rotate keys for
```
danielflookovo:
```suggestion
- `aws-username` - user name of the AWS user you want to rotate keys for
```

aws-rotate-keys/README.md
mwz: I've added a default executor.

aws-rotate-keys/README.md
mwz: That's a good point. I've removed those and I use `$CIRCLE_PROJECT_USERNAME` and `$CIRCLE_PROJECT_REPONAME` instead.

aws-rotate-keys/README.md
mwz: addressed

mwz:
Thanks for your review @danielflookovo - I've addressed all of your comments.
I've also added a new job to validate orbs (cfe6261) which can be used as a PR check, but the repo doesn't seem to be set up to build forked PRs - could we change that?

danielflookovo:
Enabling building forks is too risky given this a public repo and the secrets available in CIrcleCI. Anyone who forks could inject code into any build that used our orbs! You can disable sharing secrets to forks, but i'm not sure it's very useful then.

I've given you write access to the repo instead.

danielflookovo:
I had an idea, if you comment out the `commit` workflow from this PR i'll merge it.
Then you can open a new branch in this repo which restores it, and circleci should pick it up.

aws-rotate-keys/executor/Dockerfile
```diff
@@ -0,0 +1,3 @@
+FROM circleci/python:3-stretch
+
```
takac: Does this run as root?

mwz: it runs as `circleci` just like in he parent image

mwz:
> You can disable sharing secrets to forks, but i'm not sure it's very useful then.

I think it is still useful for running simple PR checks (which don't require secrets) like e.g. checking that the contributed orb is valid (like in my example) and making sure that the Dockerfiles are built successfully etc. These are little things which you might easily miss during code review only to find out that your pipeline fails after you've accepted and merged the PR.

danielflookovo:
It would fail the commit workflow, which requires the secrets.

mwz:
I can add filters to the `commit` workflow so it only gets run for master and I can remove publishing a dev version of the orbs as a part of that workflow. Is that something you would be happy with?

An alternative is to remove the new job which validates the orbs and reopen this PR directly in this repo instead of from a fork, as per your suggestion, but that doesn't help us with improving the workflow for any external contributors.

danielflookovo:
I want the commit workflow to run for every branch - it publishes a dev orb for testing.

mwz:
Ok, I'll close this PR and reopen it from a branch directly in this repo.

PR #32, opened 2019-06-17
Correctly publish the first version of an orb

danielflookovo:


.circleci/config.yml
```diff
@@ -67,9 +67,13 @@ jobs:
             circleci orb publish "/tmp/${NAME}_orb.yml" "ovotech/${NAME}@dev:$CIRCLE_BRANCH" --token "$CIRCLECI_TOKEN"
 
             if [ "$CIRCLE_BRANCH" = "master" ]; then
-              circleci orb source "ovotech/${NAME}" > "/tmp/${NAME}_current.yml"
+              if ! circleci orb source "ovotech/${NAME}" > "/tmp/${NAME}_current.yml"; then
+                # This is the first version
+                exec circleci orb publish promote "ovotech/${NAME}@dev:$CIRCLE_BRANCH" major --token "$CIRCLECI_PROD_TOKEN"
```
mwz: Why do you need exec here?

danielflookovo: After exec'ing, the script will no longer be running to try the `publish increment` in the next lines. It's a confusing way to do that, i'll change it.

danielflookovo: After exec'ing, the script will no longer be running to try the `publish increment` in the next lines. It's a confusing way to do that, i'll change it.

PR #33, opened 2019-06-17
Output plan to stdout

danielflookovo:


takac:
my dodgy bash looks good to me 😄

PR #34, opened 2019-06-18
Mask secrets

danielflookovo:
Add tfmask to the default executor, and use it to mask output from plan and apply.

The default resource mask pattern is extended to include kubernetes_secret and acme_certificate resources.


danielflookovo:
@JackTreble Can you test this by changing the orb version in your config to `ovotech/terraform@dev:tfmask`?

JackTreble:
@danielflookovo yep can do 👍

JackTreble:
@danielflookovo It worked [Here](https://github.com/ovotech/tardis-infra-application/pull/19)

PR #37, opened 2019-06-20
Add terraform 0.12 executor

### LINKED ISSUE #21: Update terraform orb for 0.12

danielflookovo:
With this executor you can use the orb with terraform 0.12.
Also adds a terraform_0-11 executor for symmetry.
The default executor remains and is the same as terraform_0-11.

Changed the circleci job to build multiple images and insert the digests into the orb yaml.

Closes #21


.circleci/config.yml
```diff
@@ -38,41 +38,68 @@ jobs:
...
-            docker build --tag "ovotech/${NAME}:${TAG}" \
+            for path in $ORB/executor/Dockerfile*; do
+
+              filename=$(basename $path)
+
+              if [[ "$filename" == "Dockerfile" ]]; then
+                  TAG="latest"
+                  BUILD_TAG="$(date "+%d-%m-%Y")"
+              else
+                  TAG="${filename#"Dockerfile-"}"
```
takac: I'm not sure I follow why/how there can be multiple executors which could overwrite TAG & BUILD_TAG depending on alphabetical ordering from `for path in $ORB/executor/Dockerfile*; do ...`

danielflookovo: For each dockerfile we build and push an image.
- The TAG is based on the name of the Dockerfile ('latest', '0.11' etc), and only pushed for the master branch.
- BUILD_TAG is the same but with a timestamp, only used because you need to specify something when you push.
After pushing we get the repo digest of that image and save it to a file named after the tag.

These variable are only used inside the loop.

takac: oh yes I didn't spot the indent... my bad

PR #44, opened 2019-07-05
Add jq to Terraform Dockerfiles

hayfield:
This provides a means to parse output from the AWS CLI.


hayfield:
It's used in https://github.com/ovotech/aws-network-guide

PR #52, opened 2019-07-18
Handle exit code 5 for unsupported images

skjolber:
* Support for new status code 5, which is returned when image scanning cannot be performed. With an option to fail or not to fail if that is the case.
 * Support for unexpected status codes. 
 * Delete report after each image, if present
 * Less verbose, prints less to console. Results are present as artifacts, which is sufficient.

Note; I'm not very experienced in i bash scripts.

### DIFF
--- a/clair-scanner/orb.yml
+++ b/clair-scanner/orb.yml
@@ -31,12 +31,16 @@ commands:
         type: "boolean"
         description: "Fail command when vulnerabilities at severity equal to or above the threshold are discovered"
         default: true
+      fail_on_unsupported_images:
+        type: "boolean"
+        description: "Fail command when image cannot be scanned for vulnerabilities"
+        default: true
     steps:
       - checkout
       - setup_remote_docker:
           version: 17.11.0-ce
       - run:
-          name: "vulnerability scan"
+          name: "Vulnerability scan"
           command: |
             include scan.sh
       - store_artifacts:
--- a/clair-scanner/scan.sh
+++ b/clair-scanner/scan.sh
@@ -24,7 +24,9 @@ if [ -n "<< parameters.whitelist >>" ]; then
     WHITELIST="-w /whitelist.yml"
 fi
 
-EXIT_STATUS=0
+FOUND_UNAPPROVED_VULERNERABILITIES=0
+FOUND_UNSUPPORTED_IMAGE=0
+FOUND_UNKNOWN_EXIT_STATUS=0
 
 function scan() {
     local image=$1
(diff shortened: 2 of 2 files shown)


eversC:
Hi @skjolber, thanks for your changes, and PR, and apologies for taking so long for us to review.

We'd like to make some minor changes, so I'll be pushing your fork to a branch in this repo that we can work from (that way we can keep your commits).

Some comments on the points you made:

> Support for new status code 5, which is returned when image scanning cannot be performed. With an option to fail or not to fail if that is the case.
Support for unexpected status codes.

This looks good, though some cleanup required in the bash script


> Delete report after each image, if present

We're already copying the report back out into a filename suffixed with the image name, so it seems to make sense to create a file following the same naming convention in the first place (that way we don't need to delete the report after each image scan).


> Less verbose, prints less to console. Results are present as artifacts, which is sufficient.

Could make it configurable? I know some users appreciate having the console output to view in their CI/CD jobs, for example.

skjolber:
Great to see this is making progress. Will you make the final adjustments from here on?

> > Support for new status code 5, which is returned when image scanning cannot be performed. With an option to fail or not to fail if that is the case.
> > Support for unexpected status codes.
> 
> This looks good, though some cleanup required in the bash script
> 
Good! I'll gladly admit bash scripting is something I must google every time so far.

> > Delete report after each image, if present
> 
> We're already copying the report back out into a filename suffixed with the image name, so it seems to make sense to create a file following the same naming convention in the first place (that way we don't need to delete the report after each image scan).
> 
Sounds good. 

> > Less verbose, prints less to console. Results are present as artifacts, which is sufficient.
> 
> Could make it configurable? I know some users appreciate having the console output to view in their CI/CD jobs, for example.

Yes, thats probably a good idea. We're scanning a lot of images (per job), and although its a pain to download the console output to view the end result, from time to time the scan fails (freezes), and then I'd like more console output.

skjolber:
@eversC any progress?

eversC:
Hi @skjolber, I'm almost there. There's not been much change required on top of the changes you've made, I'm just dealing with the use-cases of having console Vs. file output enabled.

My colleague @danielflookovo who can review the changes is currently away but back early next week so I'll look to get these changes finished off end of this week and discuss with him.

skjolber:
Hi @danielflookovo, any progress?

skjolber:
@eversC @danielflookovo if file output vs console is an issue, save that for the next version (its been two months now).

eversC:
Hey @skjolber 

Apologies again for this dragging on, and thanks for your patience. See #70 for all the changes (including yours). As such I'm closing this PR now

PR #53, opened 2019-07-18
:sparkles: Add tfswitch/tfenv terraform version file support

danielflookovo:
Specific terraform versions can be specified using a [tfswitch](https://warrensbox.github.io/terraform-switcher/) 
`.terraformrc` or [tfenv](https://github.com/tfutils/tfenv) `.terraform-version` file in the module path.

   This allows you to use newer terraform versions without waiting for a new orb version, and also doesn't force you to upgrade with new orb versions.

Also added tests


mwz:
Nice work!

PR #55, opened 2019-07-22
Terraform Orb 1.5.0

danielflookovo:
### Added
- A `validate` command that validates a terraform module for correctness
- A `fmt-check` command that checks terraform files in a directory are in canonical format
- A `version` command that prints the terraform version and versions of terraform plugins used
- An `in-workspace` command to execute custom steps inside an initialized terraform working directory
- The `plan` command has gained the `add_github_comment` boolean parameter to disable commenting
- Support for terraform module registry authorization using the TF_REGISTRY_HOST and TF_REGISTRY_TOKEN environment variables
- The `publish-module` orb command, for publishing a module to a terraform registry

### Changed
- `terraform-0_12` executor updated to use terraform `0.12.5` as the default
- The `GCLOUD_SERVICE_KEY` no longer needs to be base64 encoded (but still may be)
- Plans with no changes will now result in a comment on GitHub
- The GitHub comment will be updated with the outcome of the apply
- Plans with no changes will now be applied, to ensure changes to outputs are stored


terraform/publish.sh
```diff
@@ -0,0 +1,61 @@
...
+    exit 2
+fi
+
+if [[ "$VERSION_FILE_PATH" == "" ]]; then
+    echo "version_file_path parameter must be set"
+    exit 2
+fi
+
+if [[ ! -f "$VERSION_FILE_PATH" ]]; then
+  echo "Version file \"$VERSION_FILE_PATH\" doesn't exist"
```
takac: should it error here, it cats out the file below?

terraform/test/github/delete_comments.py
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/env python3
+
+import os
+import requests
+
+github_username = os.environ['GITHUB_USERNAME']
+github_token = os.environ['GITHUB_TOKEN']
+owner = os.environ['CIRCLE_PROJECT_USERNAME']
+repo = os.environ['CIRCLE_PROJECT_REPONAME']
+pr_number = os.environ.get('CIRCLE_PR_NUMBER')
```
takac: This can be `None` with `get`

terraform/publish.sh
danielflookovo: yes it should

PR #57, opened 2019-08-05
Update status when beginning plan for apply

danielflookovo:
Also fixes error updates


takac:
Can we do the same for starting a plan, as these can also take a while

danielflookovo:
Good idea, that is a bigger change though

PR #78, opened 2019-10-18
aws-rotate-keys: wait until new token usable

kelemensanyi:
aws-rotate-keys orb: waiting until new token is usable

### DIFF
--- a/aws-rotate-keys/orb.yml
+++ b/aws-rotate-keys/orb.yml
@@ -43,4 +43,19 @@ commands:
             curl --fail https://circleci.com/api/v1.1/project/github/$CIRCLE_PROJECT_USERNAME/$CIRCLE_PROJECT_REPONAME/envvar?circle-token=<< parameters.circleci-token >> \
               -X POST --header "Content-Type: application/json" \
               -d "{ \"name\":\"<< parameters.aws-secret-access-key-var >>\", \"value\":\"${new_secret_access_key}\" }" &&
-            aws iam delete-access-key --user-name << parameters.aws-username >> --access-key-id ${old_access_key_id}
+            aws iam delete-access-key --user-name << parameters.aws-username >> --access-key-id ${old_access_key_id} &&
+            echo &&
+            # Although the new token is visible on the aws API, aws does not accept it
+            # for authentication immediately.
+            AWS_ACCESS_KEY_ID=$new_access_key_id &&
+            AWS_SECRET_ACCESS_KEY=$new_secret_access_key &&
+            echo Testing new token... &&
+            while true; do
+              aws iam list-access-keys \
+                --user-name << parameters.aws-username >> \
+                --query 'AccessKeyMetadata[0].AccessKeyId' \
+                --output text 2>/dev/null && break
+              echo New token not usable yet, retrying in 10 seconds...
+              sleep 10
+            done &&
+            echo SUCCESS: New token ready to use.


kelemensanyi:
This PR is useful, because it initialises the "Last used" field of the new token, and it prevents the InfoSec alarm 1_3_unused_credentials to report it.

aws-rotate-keys/orb.yml
```diff
@@ -43,4 +43,20 @@ commands:
...
               -X POST --header "Content-Type: application/json" \
               -d "{ \"name\":\"<< parameters.aws-secret-access-key-var >>\", \"value\":\"${new_secret_access_key}\" }" &&
-            aws iam delete-access-key --user-name << parameters.aws-username >> --access-key-id ${old_access_key_id}
+            aws iam delete-access-key --user-name << parameters.aws-username >> --access-key-id ${old_access_key_id} &&
+            echo &&
+            # Although the new token is visible on the aws API, aws does not accept it
+            # for authentication immediately.
+            echo 'export AWS_ACCESS_KEY_ID=$new_access_key_id' >> $BASH_ENV &&
+            echo 'export AWS_SECRET_ACCESS_KEY=$new_secret_access_key' >> $BASH_ENV &&
+            source $BASH_ENV &&
```
mwz: Can you export the above directly without echoing them and then sourcing BASH_ENV?

aws-rotate-keys/orb.yml
```diff
@@ -43,4 +43,20 @@ commands:
...
+            echo Testing new token... &&
+            while true; do \
+              aws iam list-access-keys \
```
mwz: If the new access key is not yet available wouldn't this command exit with an error code and break the loop (since `-eo pipefail` are the default shell options on CircleCI)?

mwz:
Hey @gregorath, you should add yourself to the code owners file for this orb.

aws-rotate-keys/orb.yml
kelemensanyi: Good catch, I've fixed it.

aws-rotate-keys/orb.yml
kelemensanyi: I think
- `-o pipefail` has no affect, since we don't do piping
- but I am not sure why `-e` doesn't catch it.
I tested it in CircleCI and it works. Would you like me to dig deeper?

aws-rotate-keys/orb.yml
```diff
@@ -43,4 +43,19 @@ commands:
             curl --fail https://circleci.com/api/v1.1/project/github/$CIRCLE_PROJECT_USERNAME/$CIRCLE_PROJECT_REPONAME/envvar?circle-token=<< parameters.circleci-token >> \
               -X POST --header "Content-Type: application/json" \
               -d "{ \"name\":\"<< parameters.aws-secret-access-key-var >>\", \"value\":\"${new_secret_access_key}\" }" &&
-            aws iam delete-access-key --user-name << parameters.aws-username >> --access-key-id ${old_access_key_id}
+            aws iam delete-access-key --user-name << parameters.aws-username >> --access-key-id ${old_access_key_id} &&
+            echo &&
+            # Although the new token is visible on the aws API, aws does not accept it
+            # for authentication immediately.
+            AWS_ACCESS_KEY_ID=$new_access_key_id &&
```
gregorath: since we already make a post to update the aws variables (lines 40 and 43) do we still need to set these here?

mwz: You need to set those to make aws cli use the new access key.

aws-rotate-keys/orb.yml
mwz: LGTM, but it would be good to understand why the `-e` option doesn't have any effect for failures.

aws-rotate-keys/orb.yml
kelemensanyi: As soon as we started using the rotate command we started getting InfoSec alerts about the new keys not yet used, so this is something all users of this command face. I think it worth solving it in a central place.
Another alternative would be to duplicate this logic in all users of this orb, and have it executed immediately after this rotate command, which is
- against DRY principle,
- hard to implement, since it is difficult to access the new value of the environment variables immediately after the rotate command execution (I couldn't find a way to force CircleCI to refresh the environment variables, although it might be possible by querying the CircleCI API with curl).
Do you have any alternative solution?

gregorath: sorry probably my question was not clear.
i am saying that in lines 40 and 43  we already set AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY (or we actually set the vars  the user set) so why do we need to do here `AWS_ACCESS_KEY_ID=$new_access_key_id`?
arent the values for the aws already updated?

i am think it is good to add the call here

aws-rotate-keys/orb.yml
kelemensanyi: Here is why: https://unix.stackexchange.com/questions/312631/bash-script-with-set-e-doesnt-stop-on-command

PR #84, opened 2019-11-19
[terraform] Updated to tf 0.12.16 (with 0.12.14 and 0.12.15)

josmo:
- Updated executors to include:
    - terraform 0.12.14, 0.12.15, 0.12.16
    - gcloud 272.0.0
    - awscli 1.16.284
    - helm 2.16.1

I'm not sure the exact versioning conventions you all have. Let me know if there's anything that looks off


terraform/orb_version.txt
```diff
@@ -1 +1 @@
-ovotech/terraform@1.5.12
+ovotech/terraform@1.5.16
```
danielflookovo: Why not 1.5.13?

josmo: It seemed the version was following the TF version, so I just made it match. I'm totally down to make it 13 :)

PR #99, opened 2020-01-21
Update TF Orb Documentation

mhbrooks:
Small documentation improvements.
Minor error case improved handling.

### DIFF
--- a/terraform/README.md
+++ b/terraform/README.md
@@ -14,7 +14,12 @@ The executor named `default` is the same as `terraform-0_11`
 
 ### terraform-0_11
 
-This executor uses terraform 0.11.14 by default.
+[Dockerfile](executor/Dockerfile-0.11)
+
+This executor uses terraform 0.11.14 by default. However `tfswitch` is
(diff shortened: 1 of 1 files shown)


terraform/github.py
```diff
@@ -76,6 +76,7 @@ def __init__(self, pr_url: str):
                     self._plan = match.group(1).strip()
                     self._status = match.group(2).strip()
                     return
+        raise Exception(f'No matching plan as comment on PR {pr_url} by username {github_username}.')
```
mhbrooks: Added this as currently if you set a github email as the USERNAME env, plan works okay but apply steps fail with just: 
```
None

Exited with code exit status 1
```

danielflookovo: I have a feeling this would break the apply step, as the comment won't exist yet

danielflookovo:
```suggestion

```

mhbrooks:
Yes, good idea. I've not had chance to pick this up again, good to go ahead with the doc changes

PR #103, opened 2020-03-09
aws-rotate-keys: avoid printing AccessKeyId to std out

kelemensanyi:
Avoiding printing AccessKeyId to std out in aws-rotate-keys orb.

### DIFF
--- a/aws-rotate-keys/orb.yml
+++ b/aws-rotate-keys/orb.yml
@@ -54,7 +54,7 @@ commands:
               aws iam list-access-keys \
                 --user-name << parameters.aws-username >> \
                 --query 'AccessKeyMetadata[0].AccessKeyId' \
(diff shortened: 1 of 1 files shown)


aws-rotate-keys/orb.yml
```diff
@@ -54,7 +54,7 @@ commands:
               aws iam list-access-keys \
                 --user-name << parameters.aws-username >> \
                 --query 'AccessKeyMetadata[0].AccessKeyId' \
-                --output text 2>/dev/null && break
+                --output text >/dev/null 2>/dev/null && break
```
mwz: This can be simplified to the following:

```suggestion
                --output text 2>&1 > /dev/null && break
```

kelemensanyi:
@ovotech/production-eng would you review, merge and release this please?

PR #108, opened 2020-03-23
Load environment variables from AWS parameter store

fionamatters:


bamarch:
@fionamatters it looks like this orb is already in master https://github.com/ovotech/circleci-orbs/blob/master/aws-get-parameters/orb.yml

is this an outdated PR that needs closing, or do you want to make any changes still?

takac:
@fionamatters can we close this?

PR #115, opened 2020-04-03
fix aws-rotate-keys

bamarch:
The aws-rotate-keys repo did not exist in dockerhub

* Modify circleci config to make publishing to dockerhub optional
* Update aws-rotate-keys to use public images maintained by circleci
* Update the readme (added a badge for flair)

### DIFF
--- a/aws-rotate-keys/orb.yml
+++ b/aws-rotate-keys/orb.yml
@@ -1,11 +1,11 @@
 version: 2.1
 description: An orb to rotate AWS keys
 
+orbs:
+  aws-cli: circleci/aws-cli@1
+
 executors:
-  default:
-    description: Default executor
-    docker:
-      - image: ovotech/aws-rotate-keys:latest
+  default: aws-cli/default
 
 commands:
   rotate:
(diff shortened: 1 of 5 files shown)


aws-rotate-keys/orb.yml
```diff
@@ -1,11 +1,11 @@
...
 
+orbs:
+  aws-cli: circleci/aws-cli@1
+
 executors:
-  default:
-    description: Default executor
-    docker:
-      - image: ovotech/aws-rotate-keys:latest
+  default: aws-cli/default
```
mwz, line 8: Something to keep in mind - the `cimg/python:3.8-node` image doesn't provide the latest version of jq.

aws-rotate-keys/orb.yml
```diff
@@ -30,6 +30,10 @@ commands:
           you will use to hold this value, i.e. AWS_SECRET_ACCESS_KEY.
         default: AWS_SECRET_ACCESS_KEY
     steps:
+      - aws-cli/setup:
```
mwz, line 33: Is this potentially going to break things for those who don't use an executor and already have aws cli installed? Would it make sense to only run this step if aws cli isn't found on the path (i.e. running this orb using the default executor)?

bamarch: this step checks if aws-cli is already installed before trying to install it

next, it uses the aws-cli from path to configure the credentials

aws-rotate-keys/orb.yml
bamarch: cimg/python:3.7-node / cimg/python:3.8-node and all the next-gen CI images from circle are built on the same ubuntu base

https://github.com/CircleCI-Public/cimg-base#how-this-image-works

it provides the latest version of jq in from the Ubuntu repository

I have tested and the image works :)

mwz: yeah, all I'm saying is that the 18.04 LTS doesn't ship with the latest version of jq, but I don't think this will break anything in our case

PR #117, opened 2020-04-07
Changes for interpolation of var names in aws rotate keys

bamarch:
https://github.com/ovotech/circleci-orbs/pull/115 was supposed to just update the image being used, but it actually modified the behavior

a new feature "use << parameters.aws-secret-access-key-var >> as access credentials" does not work, because you run afoul of interpolation (which users may well want to use)

it was a misguided feature anyway so that is removed, the feature rolled back, the behavior is as before

changes in this PR:

- Stop passing strings to aws-cli/setup when it expects the circleci type `env_var_name`     
- Pin the version of the orb that is a dependency, to prevent surprises in the future

### DIFF
--- a/aws-rotate-keys/orb.yml
+++ b/aws-rotate-keys/orb.yml
@@ -31,14 +31,12 @@ commands:
         default: AWS_SECRET_ACCESS_KEY
     steps:
       - aws-cli/install:
-          # This only install the cli if it is missing from $PATH
+          # This only installs the cli if it is missing from $PATH
           version: '2'
       - run:
           name: Rotate AWS keys
           command: |
-            # Configure the credentials needed to access AWS (not using aws-cli/setup since it precludes interpolation via its usage of the env_var_name type)
-            [[ -z "$AWS_ACCESS_KEY_ID" ]] && export AWS_ACCESS_KEY_ID=<< parameters.aws-access-key-id-var >>
-            [[ -z "$AWS_SECRET_ACCESS_KEY" ]] && export AWS_SECRET_ACCESS_KEY=<< parameters.aws-secret-access-key-var >>
+            # Assumes that AWS credentials have already been configured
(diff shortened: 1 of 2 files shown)


aws-rotate-keys/orb.yml
```diff
@@ -2,7 +2,7 @@ version: 2.1
 description: An orb to rotate AWS keys
 
 orbs:
-  aws-cli: circleci/aws-cli@1
+  aws-cli: circleci/aws-cli@1.0.0
```
mwz, line 5: Technically, it should be safe to specify just the major version, as any 1.x shouldn't have any breaking changes if aws are following semver.

mwz:
I'd maybe also add an [orb_version.txt](https://github.com/ovotech/circleci-orbs/blob/master/terraform/orb_version.txt) file to this repo so we have more control over the published version, currently, it always gets incremented whenever any PR gets merged and IMHO it's quite confusing to see soo many patch versions. We should fix the version.

aws-rotate-keys/orb.yml
bamarch: it's true - was worried about the default executor changing down the line and that potentially causing issues for people

I'm also thinking about publishing proper semver versions for orbs I contribute to in this repo? In that case you would need to have downstream orbs pinned exactly, to make meaningful updates to your own version

this way if we need to update we could

bamarch:
> I'd maybe also add an [orb_version.txt](https://github.com/ovotech/circleci-orbs/blob/master/terraform/orb_version.txt) file to this repo so we have more control over the published version, currently, it always gets incremented whenever any PR gets merged and IMHO it's quite confusing to see soo many patch versions. We should fix the version.

I definitely agree - I'll check the current patch version and make a PR adding orb_version.txt

aws-rotate-keys/orb.yml
mwz: yeah, that sounds sensible - it's better to be safe than sorry 😄

PR #128, opened 2020-06-09
New orb fixups

redbaron:
- with-git-deploy-key to depend on validate orb
- tools-install orb.yaml -> orb.yml
- use prod token when creating orb on ovotech namespace


whyman:
Looks good. Lets see if I can bump the image rather than include the inline cli update.

whyman:
I've pushed a rebuilt image, can you try it?

PR #129, opened 2020-06-16
Terraform: search for backend-config and var-file in module path too

redbaron:
Backwards compatible change to search for backend config file relative to module path


whyman:
Could you not just set `parameters.backend_config_file` to have a path in it?

redbaron:
> Could you not just set `parameters.backend_config_file` to have a path in it?

> Could you not just set parameters.backend_config_file to have a path in it?

Yes, that's the current workaround, but it is counter intuitive, at least for me. `path` specifies main configuration path and rest of parameters should be relative to it, as if you do `cd <<paremeters.path>>` and rest of terraform commands run from there.

whyman:
Lets at least add something to the docs too then?

redbaron:
> Lets at least add something to the docs too then?

Added note about relative paths to  the readme

PR #132, opened 2020-06-28
Update tfswitch

dflook:
v0.8.832 of tfswitch can now also read from .terraform-version files, so we don't need to implement that ourselves.
It has also learned how to read the version from the required_version constraint in the terraform config itself.

tfswitch likes to think it runs interactively, so the output gets cleaned up a little bit.


whyman:
Looks good! Can we get a orb version bump too?

Issue #113, opened 2020-04-03
[terraform] Support lock-timeout option


steinybot:
It would be nice to have support for the `lock-timeout` option.

I have two workflows `continuous_integration` (which does a `plan`) and `continuous_deployment` (does an `apply`). The CI workflow runs on all branches. The CD workflow only runs on `master` and it is also triggered via the API from other builds which trigger a redeploy. When the build runs on `master` it runs both of these workflows concurrently and one of them will always fail as it cannot get lock because the other one has it. I need to be able to set the lock timeout so that they will retry.

bamarch:
I can see why this would be very useful, especially if your CI/CD solution does not give you the ability to run jobs exclusively, or specify concurrency etc.

All the terraform commands you're using (implicit: init, explicit: plan,apply) support this parameter:

    -lock-timeout=0s    Duration to retry a state lock.

So adding this as a parameter and passing it through sounds good!

I'll probably take a crack at this - unless you were blocked by this and published a dev version already that implements the feature? PRs happily accepted always! :D

PR #133, opened 2020-07-06
Set a lock-timeout

dflook:
This will keep retrying to obtain the state lock for 5 minutes before failing.
Current behaviour is to fail immediately, and the only sensible action is to retry it manually.

5 minutes should be long enough that it's quicker than noticing and retrying manually, but not long enough
that builds are stuck if the state lock is stale.

It's not exposed as an option, but it should address the issue seen #113

I've adapted the terraform orb for GitHub actions, and this is one of the changes I made on the way. 
Check it out: https://github.com/dflook/terraform-github-actions


whyman:
Nice, can we get a version bump?

Also, love the Github Action!

PR #134, opened 2020-07-07
Add ovorotatetest tf provider

eversC:
This adds the [ovo-rotate-test provider](https://github.com/ovotech/kafka-users-rotate-test/) that teams at OVO will need to use during their testing.


takac:
`TF_OVO_ROTATE_TEST_VERSION` maybe name it better?

I'm assuming there will be 1.0 that teams will want to use at some point? rather than having to rename later we could tee up that naming now? Not sure what the plan for this is.

Has this been tested as a dev orb? I'm reticent to push this out without thorough testing

PR #135, opened 2020-07-15
Add Versioned AWS Lambda orb

tom-g-dane:
Add an orb for the migration team AWS lambda builds.

Ran `circleci orb validate migration-lambdas/orb.yml` and `circleci orb process migration-lambdas/orb.yml` locally and all looks good.

### DIFF
--- a/versioned-lambda/orb.yml
+++ b/versioned-lambda/orb.yml
@@ -25,6 +25,10 @@ jobs:
         description: "Name of the zip file containing the lambda application code. Defaults to lambda.zip"
         type: string
         default: lambda.zip
+      vulnerability-audit:
+        description: "Whether or not to run the Snyk Scan step. Defaults to true"
+        type: boolean
+        default: true
     executor: << parameters.executor >>
     steps:
       - checkout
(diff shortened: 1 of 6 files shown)


tom-g-dane:
I did some mild formatting in some of the common files (IDE did some of it) let me know if you want me to revert this.

tom-g-dane:
Marked as draft for now as feedback from Dave Connell is to make it more generic as it may be useful for teams outside of orion migrations.

tom-g-dane:
Updated the orb to be non-migration specific

versioned-lambda/orb.yml
```diff
@@ -0,0 +1,118 @@
...
+
+            zip -r -q $ZIPFILE dist node_modules
+            aws s3 cp $ZIPFILE $S3PATH/$ZIPFILE
+
+  create-lambda-version:
+    parameters:
+      build-bucket:
+        description: "AWS S3 bucket containing lambda build artifacts"
+        type: string
+        default: ovo-orion-versioned-build
```
whyman, line 82: Should this be a default considering possible public usage?

tom-g-dane: wooops, yeah that shouldn't be a default. Will fix now thanks!

PR #137, opened 2020-07-21
Set current environment

fionamatters:
Set environmental variable (default ENVIRONMENT) from current branch. Defaults to prod for tags, nonprod for main branch and sandbox otherwise. Default name for main branch is master. Also tidied up config.yaml file and removed requirement for publish-docker-image parameter

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -152,5 +152,7 @@ workflows:
                 - rac-gcp-deploy
                 - set-current-environment
                 - terraform
+                - tools-install
                 - versioned-lambda
+                - with-git-deploy-key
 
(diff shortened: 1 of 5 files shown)


.circleci/config.yml
```diff
@@ -21,57 +21,55 @@ jobs:
...
+            docker login --username $DOCKERHUB_USERNAME --password $DOCKERHUB_PASSWORD
+
+            readonly ORB="<< parameters.path >>"
+
+            for path in $ORB/executor/Dockerfile*; do
+
+              if [ -e "$path" ]; then
+                 echo "Has docker files - continuing"
+              else
+                 echo "Has no docker files - exiting step"
```
takac, line 43: is this else required? seems like the path would always exist coming from the for loop?

fionamatters: No - it's looking to see if there are docker files - bash seems strange and will still come in here if there are no files, but in that case I think it has a null.

Basically you're looping over the result of the ls - so if there are no files the ls is null but it still gives a return for that.

PR #138, opened 2020-07-22
Lower snyk severity threshold

Samuel-West:
* Lowered snyk severity threshold to low
* Bumped version to 1.0.1


Samuel-West:
Have removed orb_version.txt as requested

PR #140, opened 2020-07-29
add avn provider releases since 1.1.4

apjm:
There are a lot of tf_aiven_versions now


eversC:
taking a look 👀

eversC:
v1.2.2 (and later) of their provider is broken for us, see https://ovoenergy.slack.com/archives/G9S51FS8P/p1596041207331600

apjm:
Oh dear. Thanks for looking At it Chris! I'll look at opening an issue in the provider repo. Setting up a call with one of their team for this week too so I'll talk about it there.

takac:
Where is this at now?

apjm:
Aiven agreed that suffix wildcards like _username_*_ are a valid construction and shouldn’t be excluded by their username validation. 
They made a change to allow suffix wildcards https://github.com/aiven/terraform-provider-aiven/pull/266 and cut a release. 
I tested the release with the aiven-as-code config. Got the same invalid non-alpha numeric username errors with the new release.
Two days ago they made another change to fix that https://github.com/aiven/terraform-provider-aiven/pull/276 
Just waiting for a release to test now.

apjm:
We can't use a lot of these versions because they identify ACL usernames with a suffix wildcard as being invalid. Will be looking to update with newer versions when that issue is resolved. I.e. #154

PR #142, opened 2020-08-11
Update version with working image

Samuel-West:
Small fix to the node12.x lambda version we are using to prevent build-stopping errors in CI.

### DIFF
--- a/versioned-lambda/orb.yml
+++ b/versioned-lambda/orb.yml
@@ -8,7 +8,7 @@ orbs:
 executors:
   lambci-node12:
     docker:
-      - image: lambci/lambda:20200721-nodejs12.x
+      - image: lambci/lambda:20200721-build-nodejs12.x
 
 jobs:
   node-test-and-package:
--- a/versioned-lambda/orb.yml
+++ b/versioned-lambda/orb.yml
@@ -8,7 +8,7 @@ orbs:
 executors:
   lambci-node12:
     docker:
-      - image: lambci/lambda:build-nodejs12.x
+      - image: lambci/lambda@sha256:c4cb47ad928cb03d149c424af779294f6638bc1ad232ecb9c3d9cdba857fb484
 
 jobs:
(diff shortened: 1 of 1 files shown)


versioned-lambda/orb.yml
```diff
@@ -8,7 +8,7 @@ orbs:
 executors:
   lambci-node12:
     docker:
-      - image: lambci/lambda:build-nodejs12.x
+      - image: lambci/lambda:20200721-nodejs12.x
```
tom-g-dane, line 11: Missing `build`:
```
- image: lambci/lambda:20200721-build-nodejs12.x
```
https://hub.docker.com/layers/lambci/lambda/20200721-build-nodejs12.x/images/sha256-c4cb47ad928cb03d149c424af779294f6638bc1ad232ecb9c3d9cdba857fb484?context=explore

Samuel-West: I just spotted that too :)

versioned-lambda/orb.yml
```diff
@@ -8,7 +8,7 @@ orbs:
 executors:
   lambci-node12:
     docker:
-      - image: lambci/lambda:build-nodejs12.x
```
takac, line 11: what's the long term fix for this?

whyman:
looks like the current version is `20200806-build-nodejs12.x` ?

Samuel-West:
The current image version has been preventing us from using sudo in our CI steps, so we've reverted back to the image version we were using prior to the issue.

versioned-lambda/orb.yml
Samuel-West: We'll have to look at running the snyk/scan step in a different container that doesn't have the issue mentioned below. Don't suppose you have any recommendations for other containers?

Samuel-West: Or potentially run the entire node-test-and-package job with a different container

PR #143, opened 2020-08-11
Bump gcloud version

redbaron:
Lets see if it helps to resolve build failure


redbaron:
Maybe original failure was intermittent, but this change builds cleanly

PR #145, opened 2020-08-13
Add npm authentication & shipit to versioned-lambda

Samuel-West:
NPM Auth:
* Adds optional step to `node-test-and-package` to authenticate with npm if a token is provided in the environment variables.
* Defaults to `true` because we assume that most projects that would use this orb would be using ovotech repositories.

Shipit:
* Adds `notify-shipit` optional step to `create-lambda-version`
* defaults to `false`

Readme updated to reflect new options.

### DIFF
--- a/versioned-lambda/README.md
+++ b/versioned-lambda/README.md
@@ -108,6 +108,7 @@ workflows:
           name: deploy-uat
           build-bucket: my-build-bucket
           lambda-function-name: cool-lambda-uat
+          context: my-uat-context
           requires:
             - test-and-package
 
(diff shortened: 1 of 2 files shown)


takac:
Would be worth updating the codeowners so you can merge stuff for this orb

tom-g-dane:
> Would be worth updating the codeowners so you can merge stuff for this orb

Hey @takac , we are in the CODEOWNERS file, I specified the team name (https://github.com/orgs/ovotech/teams/orion-migration/discussions) of which we are both members but it still requests prod eng. Is there something not right about the way we defined that?

takac:
Oh it should be in the .github dir!

takac:
pushed a fix to master, now you have approval :)

tom-g-dane:
Amazing thanks a lot @takac !

PR #147, opened 2020-08-16
Update aiven provider versions

keirlawson:


whyman:
Do we really need all of these? The docker image is already pretty beefy!

keirlawson:
Probably not, I just need 1.3.5, can remove the others

eversC:
Also worth noting the terraform aiven provider is broken for us for v1.2.2 and above. They've released a change that breaks our partially wildcarded ACLs, see #140

keirlawson:
Closing in favour of #140

PR #150, opened 2020-08-18
Snyk monitor only on master branch

Samuel-West:
Monitors when `CIRCLE_BRANCH` is equal to `master`.

This prevents us getting alerted for vulnerabilities in old/duplicate/non-existant branches. Only the most recent snapshot will be monitored.

### DIFF
--- a/versioned-lambda/orb.yml
+++ b/versioned-lambda/orb.yml
@@ -61,7 +61,7 @@ jobs:
                 command: yum install -y sudo
             - when:
                 condition:
-                  equal: [master, << pipeline.git.branch >>]
+                  equal: [master, $CIRCLE_BRANCH]
                 steps:
                   snyk/scan:
                     token-variable: SNYK_TOKEN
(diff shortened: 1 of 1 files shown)


versioned-lambda/orb.yml
```diff
@@ -59,11 +59,24 @@ jobs:
...
                 name: Sudo install
                 command: yum install -y sudo
-            - snyk/scan:
-                token-variable: SNYK_TOKEN
-                monitor-on-build: true
-                severity-threshold: low
-                project: ${CIRCLE_PROJECT_REPONAME}/${CIRCLE_BRANCH}
+            - when:
+                condition:
+                  equal: [master, $CIRCLE_BRANCH]
```
tom-g-dane, line 64: I think you need to reference `<< pipeline.git.branch >>` or whatever it was in the apollo extract repo. I don't think env vars are interpolated when it evaluates the when blocks

PR #151, opened 2020-08-19
Add TF 0.13 executor

takac:
Also removed the old aiven tf providers

### DIFF
--- a/terraform/executor/Dockerfile-0.13
+++ b/terraform/executor/Dockerfile-0.13
@@ -1,0 +1,88 @@
+FROM golang:1.12.6 AS tfmask
+
+RUN git clone https://github.com/cloudposse/tfmask.git
+RUN cd tfmask && make && make go/build
+
+FROM debian:buster-slim
+
+ARG TF_VERSIONS="0.13.0"
+ARG DEFAULT_TF_VERSION=0.13.0
+ARG GCLOUD_VERSION=279.0.0
+ARG AWSCLI_VERSION=1.17.11
+ARG HELM2_VERSION=2.16.9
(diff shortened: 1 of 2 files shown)


mattslane (approved):

whyman:
Do we need to bundle any executors? Perhaps we could make them available somewhere to allow them to be used via the new "Automatic installation of third-party providers" functionality?

takac:
If you mean "providers" rather than "executors" I think that's a good idea! Wasn't aware of that feature.

takac:
@whyman there might be an issue with that.. do Aiven host a terraform registry? The auto-install requires a true registry rather than just an s3 bucket from what I've read

whyman:
I would hope that Aiven would add their provider to the hashicorp provider registry.

It looks like you can publish providers directly to the hashicorp registry (https://registry.terraform.io/publish/provider).

So initally perhaps we could upload an `ovotech/aiven` provider, until the `aiven/aiven` provider is available.

takac:
Someone else is already doing this 🤦 https://registry.terraform.io/providers/juniorz/aiven/latest

Is the `kafka-users` provider hosted in our terraform registry (and is it accessible)?

Maybe we revisit this when these are in a registry

takac:
I asked Aiven about getting their provider in the official registry

> Yes, we are currently are working on that! There is a process that we must follow to become HashiCorp approved Terraform Provider and our Terraform team is working on making the required changes to existing one. We are expecting to have it available after Aiven Terraform Provider v2 release. I will add your contact information and we will let you know once the process is complete.

takac:
Lets cut another release when we make that change to remove those providers

PR #153, opened 2020-08-19
create slim 0.12 version

takac:


terraform/executor/Dockerfile-0.12-slim
```diff
@@ -0,0 +1,88 @@
...
+      && unzip ovo.zip -d /root/.terraform.d/plugins \
+      && rm ovo.zip; \
+    done
+
+RUN curl -fsL https://dl.google.com/dl/cloudsdk/channels/rapid/downloads/google-cloud-sdk-${GCLOUD_VERSION}-linux-x86_64.tar.gz -o /opt/google-cloud-sdk.tar.gz \
+ && cd /opt; tar -xf google-cloud-sdk.tar.gz \
+ && rm google-cloud-sdk.tar.gz
+ENV PATH "/opt/google-cloud-sdk/bin:$PATH"
+
+RUN pip3 install awscli==$AWSCLI_VERSION
```
chris-brindley, line 83: If we disable the pip cache here we can save a few MB

chris-brindley (commented):
I guess the intention of a slim image is a smaller filesize but this still results in a 1.3G image. Is there any more scope to reduce the number of utilities in it?

takac:
According to dockerhub it's now ~470mb whereas non-slim is ~1.3gb

chris-brindley:
Yeah seems to be a difference between how dockerhub reports image sizes and how they are reported locally, where I checked it

chris-brindley (approved):

terraform/executor/Dockerfile-0.12-slim
```diff
@@ -0,0 +1,88 @@
...
+    python3 \
+    python3-requests \
+    python3-pip \
+    wget \
+ && rm -rf /var/lib/apt/lists/*
+
+RUN curl -fsL https://github.com/warrensbox/terraform-switcher/releases/download/${TFSWITCH_VERSION}/terraform-switcher_${TFSWITCH_VERSION}_linux_amd64.tar.gz -o tfswitch.tar.gz \
+ && tar -xvf tfswitch.tar.gz \
+ && mv tfswitch /usr/local/bin \
+ && rm -rf tfswitch \
```
oarmstrong-ovo, line 43: I think you'll also want to delete `tfswitch.tar.gz` to save a few more bytes (~9M).

oarmstrong-ovo: Correction: this line should be `tfswitch.tar.gz`.

takac: looks like it should be
```
rm -rf tfswitch.tar.gz
```
rather than `tfswitch`

takac:
369.66 MB this run

takac:
192mb compressed 
https://hub.docker.com/layers/ovotech/terraform/0.12-slim_19-08-2020/images/sha256-60a053a6b20a09dfaf7d3ce7750254f8eed2f8629e8cba92607c109ecadd86fb?context=explore

PR #154, opened 2020-08-27
upgrade aiven provider version for slim

apjm:
I understand the slim executor should have the latest version.
Aiven worked out the ACL username validation issues and we tested
that locally just running tf plan for the kafka-uat config in
aiven as code.
Aiven have also changed the structure of the filenames of their
release assets. Made some changes to executor to accommodate

After this I'll look at changing the executor for the uat jobs in aiven as
code over to using the slim executor. Not going ahead and updating 
the executor all the aac jobs use is just out of caution around 
the ACL username validation issue in particular and the fact that we’ve
not updated the version we’re using since February so there is some 
risk from the unexpected. There is a major version bump too.

### DIFF
--- a/terraform/executor/Dockerfile-0.12-slim
+++ b/terraform/executor/Dockerfile-0.12-slim
@@ -12,7 +12,7 @@ ARG AWSCLI_VERSION=1.17.11
 ARG HELM2_VERSION=2.16.9
 ARG HELM3_VERSION=3.0.3
 ARG TFSWITCH_VERSION=0.8.832
-ARG TF_AIVEN_VERSIONS="1.1.4"
+ARG TF_AIVEN_VERSIONS="2.0.1"
 ARG TF_OVO_VERSIONS="1.0.0"
 
 # Terraform environment variables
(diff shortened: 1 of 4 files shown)


terraform/executor/Dockerfile-0.12-slim
```diff
@@ -48,9 +48,10 @@ RUN mkdir -p $TF_PLUGIN_CACHE_DIR
 
 RUN mkdir -p /root/.terraform.d/plugins \
  && for TF_AIVEN_VERSION in $TF_AIVEN_VERSIONS; do \
-      curl -f -L "https://github.com/aiven/terraform-provider-aiven/releases/download/v${TF_AIVEN_VERSION}/terraform-provider-aiven-linux-amd64_v${TF_AIVEN_VERSION}" \
-        -o /root/.terraform.d/plugins/terraform-provider-aiven_v${TF_AIVEN_VERSION} \
-      && chmod +x /root/.terraform.d/plugins/terraform-provider-aiven_v${TF_AIVEN_VERSION}; \
+      curl -f -L "https://github.com/aiven/terraform-provider-aiven/releases/download/v${TF_AIVEN_VERSION}/terraform-provider-aiven_{TF_AIVEN_VERSION}_linux_amd64.zip" -o aiven-provider.zip \
```
eversC, line 51: I believe this is missing a `$` from the 2nd occurrence of `TF_AIVEN_VERSION`

eversC (changes requested):

terraform/executor/Dockerfile-0.12-slim
apjm: Bah. Thank you Chris.

eversC (approved):

takac (changes requested):
bump orb version

apjm:
> bump orb version

Righto. I had a go at the Changelog and readme too while I was there.

takac (approved):
lgtm

PR #161, opened 2020-09-29
add new aiven provider versions

bamarch:
latest releases:
https://github.com/aiven/terraform-provider-aiven/releases

see slack thread:
https://ovoenergy.slack.com/archives/C7V2SN09X/p1601378202029000?thread_ts=1601374075.025000&cid=C7V2SN09X

PR #162, opened 2020-09-29
TF Orb: add new TF 0.13.x vers and Aiven providers

takac:


terraform/executor/Dockerfile-0.13
```diff
@@ -5,14 +5,14 @@ RUN cd tfmask && make && make go/build
...
-ARG TF_VERSIONS="0.13.0 0.13.1 0.13.2"
+ARG TF_VERSIONS="0.13.0 0.13.1 0.13.2 0.13.3"
 ARG DEFAULT_TF_VERSION=0.13.2
 ARG GCLOUD_VERSION=279.0.0
 ARG AWSCLI_VERSION=1.18.134
 ARG HELM2_VERSION=2.16.10
 ARG HELM3_VERSION=3.3.1
 ARG TFSWITCH_VERSION=0.8.832
-ARG TF_AIVEN_VERSIONS="1.0.19 1.0.20 1.1.0 1.1.1 1.1.2 1.1.3 1.1.4"
+ARG TF_AIVEN_VERSIONS="2.0.6"
```
bamarch, line 15: this will force users of `0.13` to upgrade to `v2`

as per #161, could this be

```
ARG TF_AIVEN_VERSIONS="1.0.19 1.0.20 1.1.0 1.1.1 1.1.2 1.1.3 1.1.4 1.3.5"
```

or 

```
ARG TF_AIVEN_VERSIONS="1.3.5"
```

i.e. because is using 0.13 they can download v2 via terraform 0.13 support

terraform/executor/Dockerfile-0.12
```diff
@@ -12,7 +12,7 @@ ARG AWSCLI_VERSION=1.18.134
 ARG HELM2_VERSION=2.16.10
 ARG HELM3_VERSION=3.3.1
 ARG TFSWITCH_VERSION=0.8.832
-ARG TF_AIVEN_VERSIONS="1.0.19 1.0.20 1.1.0 1.1.1 1.1.2 1.1.3 1.1.4"
+ARG TF_AIVEN_VERSIONS="1.0.19 1.0.20 1.1.0 1.1.1 1.1.2 1.1.3 1.1.4 2.0.1 2.0.5 2.0.6"
```
bamarch, line 15: could this also include v1.3.5, the latest v1 provider? 

we know `v1.1.4` breaks with the new uat config, but `v1.3.5` may work

are there breaking syntax changes between 1 and 2?

bamarch:
Worth also updating `terraform/executor/Dockerfile-0.11` since we know it's broken in the current state (if using the aiven provider)

I see 14 uses by ovotech:
https://github.com/search?q=org%3Aovotech+terraform-0_11&type=code

Suggest:

```
ARG TF_AIVEN_VERSIONS="1.0.19 1.0.20 1.1.0 1.1.1 1.1.2 1.1.3 1.1.4 1.3.5 2.0.6"
```

bamarch (changes requested):
see comments

terraform/executor/Dockerfile-0.13
takac: I don't expect anyone to be using this provider tbh and we should be using this version rather than the old one.

bamarch: I can't see anyone in ovo using this image currently, so it's a moot point

https://github.com/search?q=org%3Aovotech+terraform-0_13&type=code

i'll resolve this convo

terraform/executor/Dockerfile-0.12
bamarch: resolved in latest push

terraform/executor/Dockerfile-0.13
bamarch: seen the latest change, looks like the best way to go imo

bamarch:
> Worth also updating `terraform/executor/Dockerfile-0.11` since we know it's broken in the current state (if using the aiven provider)
> 
> I see 14 uses by ovotech:
> https://github.com/search?q=org%3Aovotech+terraform-0_11&type=code
> 
> Suggest:
> 
> ```
> ARG TF_AIVEN_VERSIONS="1.3.5"
> ```
> 
> Because v2 of the aiven provider does not support < terraform 0.12
> ([aiven/terraform-provider-aiven@b20d4e7](https://github.com/aiven/terraform-provider-aiven/commit/b20d4e730a8cd999689ad82e0f5738d8748ceefe))

boost / migrations / payg are the only people using 0.11 image:

https://github.com/ovotech/boost-bit-workflows/blob/4698617cc9d2dda3a256ccfeadd3a01fd1eef910/.circleci/config.yml
https://github.com/ovotech/orion-migration-statement-load/blob/d07cbaf7d6735d5874bb77cb98b78507b80ab620/.circleci/config.yml
https://github.com/ovotech/payg-account-service/blob/daa2d6afb19bcfbc3546e46a3641947a53ea0548/.circleci/config.yml

of those, only boost uses the aiven provider with it

so maybe give boost a heads up, but I'll accept the PR as is now

bamarch (approved):

PR #165, opened 2020-10-05
 Implement an orb for setting up an SSH tunnel for remote deployments

chongyangshi:
This is an orb for running a [`sshuttle`](https://sshuttle.readthedocs.io/) proxy in the background, which is a routed SSH tunnel to a remote network, through a bastion server or an ingress load balancer. This establishes a bridge to the remote network for performing deployments against private endpoints. The proxy server should be secured through various standard precautions for servers used as SSH proxy, including what internal network destination can be reached from the proxy server itself. 

For more details on the use-case -- which I'm not keen to discuss in a public repository -- happy to have a chat on Slack and point you to the relevant places.


ssh-proxy/orb.yml
```diff
@@ -0,0 +1,72 @@
...
+
+    steps:
+      - add_ssh_keys:
+          fingerprints:
+            - << parameters.private_key_fingerprint >>
+      - run:
+          name: Install additional packages required
+          shell: /bin/bash
+          command: |
+            sudo apt update && sudo apt install -y sshuttle
```
chongyangshi, line 39: This is a Ubuntu package in the official repository: https://packages.ubuntu.com/xenial/sshuttle

Not installing this in a Dockerfile because this proxy needs to run on a VM in CircleCI instead of a container anyway.

chongyangshi:
Re-opening without fork.

PR #166, opened 2020-10-05
Implement an orb for setting up an SSH tunnel

chongyangshi:
This is an orb for running a sshuttle proxy in the background, which is a routed SSH tunnel to a remote network, through a bastion server or an ingress load balancer. This establishes a bridge to the remote network for performing deployments against private endpoints. The proxy server should be secured through various standard precautions for servers used as SSH proxy, including what internal network destination can be reached from the proxy server itself.

For more details on the use-case -- which I'm not keen to discuss in a public repository -- can be found here: https://ovoenergy.slack.com/archives/CC3CFQY0M/p1601913299163100


redbaron:
I wonder if `gcloud compute ssh` to one of existing VMs can be used instead of bastion hosts? It uses Identity aware proxy under the hood is capable of establishing arbitrary TCP tunnels

chongyangshi:
@redbaron I tried to set IAP up with a GCP NLB before, couldn't get health check to work for some reason, but with enough persistence there's probably a way to make it work. Does't have an equivalent in AWS though.

ssh-proxy/orb.yml
```diff
@@ -63,7 +63,7 @@ commands:
     steps:
       - run:
           name: Release all sshuttle sessions, which will kill forked ssh sessions as well
-          command: sudo sh -c "ps a | grep -- 'sshuttle' | grep -v 'grep' | awk '{print $2}' | xargs kill"
+          command: sudo sh -c "ps au | grep -- 'sshuttle' | grep -v 'grep' | awk '{print $2}' | xargs kill"
```
takac, line 66: `sudo pkill sshuttle`?

chongyangshi: Yeah, that's much cleaner, I'll change that

connelldave (approved):

chongyangshi:
https://ovoenergy.slack.com/archives/CC3CFQY0M/p1602238519213600?thread_ts=1601913299.163100&cid=CC3CFQY0M

PR #167, opened 2020-10-05
Terraform Orb 2.0.0

takac:
- Make 0.13 the default TF version
- Remove 0.11 executor
- Remove 0.12 "heavy" executor, (keep slim version)


bamarch:
Remove 0.11 executor - seems fine - I think we sleuthed in another PR that there is perhaps just one old boost service using it?

Remove 0.12 "heavy" executor, (keep slim version) - this is a bit more controversial, people might have been depending on things in that image (e.g. `wget`) that aren't available in the slim one, so it's not 100pct a simple replacement

But I suppose you just break deployments not running apps, and can revert this change, so maybe go ahead with it. Give teams a heads up though?

whyman:
Isnt this `2.0.0`?

I would add `wget` to the slim image and then...

eversC:
closing in favour of #264

PR #158, opened 2020-09-17
Add HCL syntax highlighting for Terraform plan GitHub comment

gordonkeogh:

PR #175, opened 2020-10-13
Revert HCL highlighting

eversC:
Reverts the changes made in #158, I saw a failed job when using the new orb version with comments on a PR from an old version. I've still not tracked any specific errors down, but safer to revert for now I think..


enicholson125 (approved):
lgtm

PR #177, opened 2020-10-16
Add hcl formatting

eversC:
- created `terraform/comment_util.py` with functions for regex matching and creation of the comment that's put onto PRs, and updated `terraform/github.py` to use new functions
- created `terraform/test_comment_util.py` with some tests in
- updated the `terraform/plan.sh` and `terraform/apply.sh` to copy new util python file to `/tmp`

I tested the devorb, and both plan and apply were successful (with HCL highlighting).

I haven't been able to test the case of a plan being done on old version, and apply being done on new version, but I think my unit tests cover that.

Credit to @gordonkeogh for the initial work on this; this PR is just to fix the changes on #158

### DIFF
--- a/terraform/CHANGELOG.md
+++ b/terraform/CHANGELOG.md
@@ -3,6 +3,10 @@ All notable changes to the orb will be documented in this file.
 Orbs are immutable, some orb versions with no significant changes are
 not listed
 
+## ovotech/terraform@1.7.8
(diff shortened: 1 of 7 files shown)


eversC:
Oh, I'll push another commit to dump the release version...

whyman (approved):

PR #181, opened 2020-10-21
Change save test results

ChrisFord-Ovo:
Add ability to build libs separately to resolve issue with same files being created in separate workspaces


mattslane (approved):
LGTM

PR #184, opened 2020-10-26
aws-rotate-keys: add parameter checks

clarencedglee:
Added checks to the parameters on aws-rotate-keys so that users get an error message if any of them are empty.

It's possible to pass a variable in that's not defined, and it's not obvious to debug this. In a current project, we lost quite a bit of time to find out the `CIRCLECI_TOKEN` was not automatically provided as an environment variable the way other `CIRCLECI_XXX` variables were.

This change aims to make that more obvious to future users.

### DIFF
--- a/aws-rotate-keys/orb.yml
+++ b/aws-rotate-keys/orb.yml
@@ -36,6 +36,29 @@ commands:
       - run:
           name: Rotate AWS keys
           command: |
+            # Check parameters have been provided
+            has_missing_parameters=false
+            missing_param_message="Error: Please provide a non-empty string for parameters:"
+            if [ -z << parameters.aws-username >> ]; then
+              has_missing_parameters=true
+              missing_param_message="$missing_param_message\n- aws-username"
+            fi
+            if [ -z << parameters.circleci-token >> ]; then
+              has_missing_parameters=true
+              missing_param_message="$missing_param_message\n- circleci-token"
+            fi
+            if [ -z << parameters.aws-access-key-id-var >> ]; then
+              has_missing_parameters=true
+              missing_param_message="$missing_param_message\n- aws-access-key-id-var"
+            fi
+            if [ -z << parameters.aws-secret-access-key-var >> ]; then
+              has_missing_parameters=true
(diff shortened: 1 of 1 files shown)


aws-rotate-keys/orb.yml
```diff
@@ -36,6 +36,29 @@ commands:
...
+              missing_param_message="$missing_param_message\n- aws-access-key-id-var"
+            fi
+            if [ -z << parameters.aws-secret-access-key-var >> ]; then
+              has_missing_parameters=true
+              missing_param_message="$missing_param_message\n- aws-secret-access-key-var"
+            fi
+            if $has_missing_parameters; then
+              echo -e $missing_param_message
+              exit 1
+            fi
```
clarencedglee, line 61: Would've been nice to used a `for` looping, but didn't see a way since the shell script would be unaware of CircleCI's variables in `<< parameters.aws-secret-access-key-var >>`.

clarencedglee:
This orb didn't have a `orb_version.txt`, does it need one now?

aws-rotate-keys/orb.yml
```diff
@@ -36,6 +36,29 @@ commands:
...
+              missing_param_message="$missing_param_message\n- circleci-token"
+            fi
+            if [ -z << parameters.aws-access-key-id-var >> ]; then
+              has_missing_parameters=true
+              missing_param_message="$missing_param_message\n- aws-access-key-id-var"
+            fi
+            if [ -z << parameters.aws-secret-access-key-var >> ]; then
+              has_missing_parameters=true
+              missing_param_message="$missing_param_message\n- aws-secret-access-key-var"
+            fi
```
mwz, line 57: These two default to env variables, so maybe you could mention the name of the env variables in those error messages, e.g. something like this? `- aws-access-key-id-var or AWS_ACCESS_KEY_ID environment variable` etc.

mwz:
> This orb didn't have a `orb_version.txt`, does it need one now?

Yes, it would be a good idea to prevent it from being published every single time the pipeline is run.

mwz:
closing as stale

PR #197, opened 2020-11-11
rac-gcp-deploy: Bump build image to openjdk11

redbaron:


rmorais (approved):

rac-gcp-deploy/executor/Dockerfile
```diff
@@ -4,7 +4,7 @@ FROM google/cloud-sdk:${GCLOUD_SDK_VERSION:-304.0.0-alpine}
 ARG SBT_VERSION
 ENV SBT_VERSION ${SBT_VERSION:-1.3.13}
 
-RUN apk --update add openjdk8
+RUN apk --update add openjdk11
```
harrylaou, line 7: Does this upgrade jdk to 11 for all projects? 
Sometimes upgrading to 11 , removes some classes from classpath. 
It is easily fixed. But we need to monitor all pods for exceptions. 
Is there a way to use stern (or similar) for all namespaces?

PR #198, opened 2020-11-16
Use Snyk orb

rmorais:


rac-gcp-deploy/orb.yml
```diff
@@ -288,6 +281,88 @@ jobs:
...
+    executor: <<parameters.executor>>
+    steps:
+      - attach_workspace:
+          at: <<parameters.workspace-dir>>
+      - load_cache:
+          cache-suffix: <<parameters.cache-suffix>>
+      - run: *sbt-dependency-graph
+      - snyk/scan:
+          project: $CIRCLE_PROJECT_REPONAME
+          fail-on-issues: true
```
jensraaby, line 307: Will there be a way to override the fail-on-issues if we get stuck with a particular dependency?
(It's a bit of a risk though as you'd be allowing broken windows...)

rmorais: At the moment there isn't any override.

rac-gcp-deploy/orb.yml
```diff
@@ -288,6 +281,115 @@ jobs:
...
+          at: <<parameters.workspace-dir>>
+      - load_cache:
+          cache-suffix: <<parameters.cache-suffix>>
+      - run: *sbt-dependency-graph
+      - snyk/scan:
+          project: $CIRCLE_PROJECT_REPONAME
+          fail-on-issues: false
+          monitor-on-build: true
+
+  pr_comment_snyk:
```
redbaron, line 337: IMHO these should be commands, not jobs. Jobs require start of the new executor which increases contention on CircleCI and makes builds longer

rac-gcp-deploy/orb.yml
```diff
@@ -288,6 +281,115 @@ jobs:
...
+      - load_cache:
+          cache-suffix: <<parameters.cache-suffix>>
+      - run: *sbt-dependency-graph
+      - run:
+          name: Post Snyk report to GitHub PR
+          command: |
+            PR_COMMENT_TITLE="SNYK REPORT"
+
+            comment_pr () {
+              local pr_body="$PR_COMMENT_TITLE"$'\n```PowerShell\n'"$snyk_report"$'\n```'
```
redbaron, line 364: what is `$` in `$'\n` ?

rmorais: some bash syntax to allow the correct handling of `\n` in strings

rac-gcp-deploy/orb.yml
```diff
@@ -288,6 +281,115 @@ jobs:
...
+    executor: <<parameters.executor>>
+    steps:
+      - attach_workspace:
+          at: <<parameters.workspace-dir>>
+      - load_cache:
+          cache-suffix: <<parameters.cache-suffix>>
+      - run: *sbt-dependency-graph
+      - run:
+          name: Post Snyk report to GitHub PR
+          command: |
```
tcsquires, line 360: Can this (should this) Bash be extracted into a `.sh` script file that the Orb invokes?

rac-gcp-deploy/orb.yml
rmorais: By these, you are referring to all of the snyk jobs or just the `pr_comment_snyk`one?

rac-gcp-deploy/orb.yml
redbaron: TIL

rac-gcp-deploy/orb.yml
rmorais: I think it can. I will have a look at it

rac-gcp-deploy/orb.yml
rmorais: I can make the snyk_monitor a job and include it as a step of the release job.
As for the other two jobs, I don't think they be made simple commands and included as a step on the existing jobs. These should be run only for PR's but we don't any specific jobs only for PRs.
I happy to hear any suggestions you might have.

rac-gcp-deploy/orb.yml
redbaron: not possible AFAIK :( only option is inline orb or bake it into an docker image orb must be used with

rac-gcp-deploy/orb.yml
```diff
@@ -340,37 +340,7 @@ jobs:
...
-              -u $GITHUB_USERNAME:$GITHUB_TOKEN | jq -r '.[] | select(.user.login == "'$GITHUB_USERNAME'") | select(.body | startswith("'"$PR_COMMENT_TITLE"'")) | .url')
-
-            if [ -z "$existing_pr_comment" ]; then
-              echo "Adding new comment"
-              comment_pr $pr_comment_url
-            else
-              echo "Updating existing comment"
-              comment_pr $existing_pr_comment
-            fi
+            include github_comment_pr.sh
```
redbaron, line 343: what is this magic?

rmorais: not sure if it works but other orbs in this repo, inline scripts like this.
If and when CircleCi is functioning properly again, I will tell you if it works

redbaron: OMG, orbs are preprocessed before being published: https://github.com/ovotech/circleci-orbs/blob/7ef7e85c8d505f2ff01604483e8f11f0d9a9c8aa/tools/include.py

rmorais: so this actually seems to work :)
the external script was inlined in the compiled configuration

rmorais: I guess that preprocessor is doing the same as https://circleci.com/docs/2.0/using-orbs/#file-include-syntax

redbaron: Nice find!

redbaron (approved):

sophiecollard (approved):
Approving based on your demo of this 😄

PR #199, opened 2020-11-17
Added new oot-deploy orb.

kelveden:
This orb is designed for OOT to use in deploying services via gitops using ArgoCD. However, it has been documented in such a way that it should be reusable by other folks (within the limits of the interpolation and expected gitops file structure anyway).


takac (commented):
CODEOWNERS file update plz 👍

kelveden:
> CODEOWNERS file update plz +1

Ah yeah good shout; thanks @takac

takac (approved):

takac (approved):

PR #209, opened 2020-12-07
Add slack notification

ChrisFord-Ovo:


jaws-journey-deploy/README.md
```diff
@@ -155,9 +163,21 @@ Does not require any parameters passed through.  However it does require that yo
...
+This step will send a slack notification when a build step fails.  This is to increase visibility of failures to the development team.
+
+To run the slack notification we implemented the steps in the [developers setup guide](https://github.com/CircleCI-Public/slack-orb/wiki/Setup#4-create-a-context-on-circleci).  For CircleCi you just need to make sure you are passing the following environment variables
+- SLACK_ACCESS_TOKEN
+- SLACK_DEFAULT_CHANNEL - to get the channel Id it is easier to open slack in a web browser as you can see the channel Id in the url  
+
+#### Example notification
+![Notification Sample](./images/notification_sample.PNG)
+
 ## Example Usage
```
gcollins-spark, line 181: Maybe add the new parameter to the example

jaws-journey-deploy/orb.yml
```diff
@@ -49,6 +53,11 @@ aliases:
     var_file: << parameters.environment >>.tfvars
     label: << parameters.environment >>
 
+  slack-notification: &slack-notification
+    event: fail
+    mentions: '@jaws_support'
```
gcollins-spark, line 58: It is "@jaws-support"

gcollins-spark: Should this also come from config though?

gcollins-spark: I think OSO are using this orb for JAWS API and they might want to notify their support

jaws-journey-deploy/README.md
ChrisFord-Ovo: removed parameter

jaws-journey-deploy/orb.yml
ChrisFord-Ovo: yeah its jaws-support.  Passing it into the orb, might be annoying, only option I can think of would be as a env var which is also a bit rubbish, and wouldnt help with OSO example (as they use the JAWS contexts).  Will have a think, but right now I think having it set to jaws-support regardless is fine - OSO can look in the alert channels for now

andrewinci (approved):

takac (approved):

jaws-journey-deploy/README.md
```diff
@@ -32,10 +34,9 @@ Does not require any parameters to be provided
 
 **Description**
 
-This job performs two steps for the journey code bases.  
+This job performs three steps for the journey code bases.
```
fionamatters, line 37: Is the third step a secret?

jaws-journey-deploy/README.md
```diff
@@ -18,6 +18,8 @@ Available Jobs
 * [notify-shipit](#notify-shipit)
 * [run-automation-test](#run-automation-test)
 
+All the available jobs apart from checkout-code will notify of failure via a slack message.  For more info go to [slack-notification](#slack-notification) section
```
fionamatters, line 21: Should maybe say only in prod and nonprod?

ChrisFord-Ovo: technically it could be any environment - it is dependant on what setting SLACK_INTEGRATION_ENABLED has

jaws-journey-deploy/orb.yml
```diff
@@ -49,6 +50,11 @@ aliases:
     var_file: << parameters.environment >>.tfvars
     label: << parameters.environment >>
 
+  slack-notification: &slack-notification
+    event: fail
+    mentions: '@jaws-support'
```
fionamatters, line 55: Wondering if the mentions should be a passed variable - just in case we have multiple support teams in the future? - Probably not though and it would make updating hard if we changed the link.

jaws-journey-deploy/README.md
fionamatters: I should really read the whole thing before I post

jaws-journey-deploy/README.md
ChrisFord-Ovo: updated

fionamatters (approved):
Other than a couple of bits that looks really good.

jaws-journey-deploy/orb.yml
ChrisFord-Ovo: might be something to look at in the future

gcollins-spark (approved):
👍

PR #217, opened 2020-12-29
Create release

ChrisFord-Ovo:
Add semantic release to jaws deployment orb so that we can auto create releases, update changelog etc


rekaelek-ovo (approved):
lgtm

PR #221, opened 2021-01-29
Use chdir in 0.14 terraform

enicholson125:
Terraform 0.14 introduces the -chdir parameter to run terraform in a specific directory. The previous mechanism of specifying the terraform config directory path at the end of the terraform command is deprecated but still works with 0.14 however it does not pick up the provider lock file in the config directory. Instead the provider lock file in the directory the command is being run from is used.

I've added this as an optional parameter by introducing the `use_chdir` parameter which defaults to false. The reason for doing this is that there is a subtle difference in the -var-file behaviour in the orb with chdir. -chdir expects the variable files to always be specified relative to the directory containing the Terraform config, whereas the orb currently allows you to specify the variable files either relative to the Terraform config or from the directory where terraform is running. Implementing -chdir for all users on terraform >= 0.14 would result in variable file errors for any users who are specifying their variable file paths from the running directory, so an opt-in system seemed better.

### DIFF
--- a/terraform/apply.sh
+++ b/terraform/apply.sh
@@ -46,7 +46,7 @@ update_status "Applying plan in CircleCI Job [${CIRCLE_JOB}](${CIRCLE_BUILD_URL}
 exec 3>&1
 
 set +e
-terraform "$chdir" plan -input=false -no-color -detailed-exitcode -lock-timeout=300s -out=plan.out $PLAN_ARGS "$config_path" \
+terraform "$chdir" plan -input=false -no-color -detailed-exitcode -lock-timeout=300s -out=plan.out $PLAN_ARGS $config_path \
     | $TFMASK \
     | tee /dev/fd/3 \
     | $COMPACT_PLAN \
--- a/terraform/apply.sh
+++ b/terraform/apply.sh
@@ -46,7 +46,7 @@ update_status "Applying plan in CircleCI Job [${CIRCLE_JOB}](${CIRCLE_BUILD_URL}
 exec 3>&1
 
 set +e
-terraform $chdir plan -input=false -no-color -detailed-exitcode -lock-timeout=300s -out=plan.out $PLAN_ARGS "$config_path" \
+terraform "$chdir" plan -input=false -no-color -detailed-exitcode -lock-timeout=300s -out=plan.out $PLAN_ARGS "$config_path" \
     | $TFMASK \
     | tee /dev/fd/3 \
     | $COMPACT_PLAN \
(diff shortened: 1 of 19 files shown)


terraform/apply.sh
```diff
@@ -46,7 +46,7 @@ update_status "Applying plan in CircleCI Job [${CIRCLE_JOB}](${CIRCLE_BUILD_URL}
 exec 3>&1
 
 set +e
-terraform plan -input=false -no-color -detailed-exitcode -lock-timeout=300s -out=plan.out $PLAN_ARGS "$module_path" \
+terraform $chdir plan -input=false -no-color -detailed-exitcode -lock-timeout=300s -out=plan.out $PLAN_ARGS "$config_path" \
```
chris-brindley, line 49: Missing quotation marks around `$chdir` here

chris-brindley:
The only other comment I have is whether we need to include the `--chdir` flag on the `output.sh` command as well. But I don't know enough about the orb to say one way or the other

enicholson125:
> The only other comment I have is whether we need to include the `--chdir` flag on the `output.sh` command as well. But I don't know enough about the orb to say one way or the other

Good point, thank you - I think that will need chdir but I'll just run a few experiments locally as I can't see how it's working at present.

enicholson125:
> > The only other comment I have is whether we need to include the `--chdir` flag on the `output.sh` command as well. But I don't know enough about the orb to say one way or the other
> 
> Good point, thank you - I think that will need chdir but I'll just run a few experiments locally as I can't see how it's working at present.

The `terraform output` command currently only works if you run it within the directory containing the terraform output config - so it won't currently be working for users who are specifying a different directory in the path parameter. I've added the -chdir parameter so that it will now work as expected for users who specify a path parameter and are using a terraform version >= 0.14, but the existing behaviour won't change for terraform < 0.14 (as `terraform output` doesn't support terminal config directory parameters)

eversC (approved):

eversC:
Please could we also update `terraform/CHANGELOG.md` and `terraform/orb_version.txt` accordingly?

eversC (changes requested):
Please could we also update terraform/CHANGELOG.md and terraform/orb_version.txt accordingly?

terraform/validate.sh
```diff
@@ -1,8 +1,8 @@
 rm -rf .terraform
-terraform init -input=false -backend=false -no-color "$module_path"
+terraform "$chdir" init -input=false -backend=false -no-color "$config_path"
```
eversC, line 2: it seems as though this is coming out as:

```bash
terraform -chdir=cluster init -input=false -backend=false -no-color ''
```

and the single quotes at the end are breaking it.

eversC (changes requested):

eversC (approved):

eversC:
seems like your dev testing was successful @enicholson125 , as was the brief spell I just conducted

lgtm 🎆

PR #223, opened 2021-01-30
R&C orb updates

rmorais:


rac-gcp-deploy/orb.yml
```diff
@@ -467,62 +489,3 @@ jobs:
...
-        default: ""
-        type: string
-    executor: <<parameters.executor>>
-    steps:
-      - restore_workspace:
-          to: <<parameters.workspace-dir>>
-      - load_cache:
-          cache-suffix: <<parameters.cache-suffix>>
-      - gcloud_auth: *gcloud_auth_prd
-      - k8s_auth:
```
lkrystkowiak-ovoenergy, line 520: can this command be removed now? seems like it's not used anymore

rac-gcp-deploy/executor/Dockerfile
```diff
@@ -2,27 +2,15 @@ ARG GCLOUD_SDK_VERSION
 FROM google/cloud-sdk:${GCLOUD_SDK_VERSION:-304.0.0-alpine}
 
 ARG SBT_VERSION
-ENV SBT_VERSION ${SBT_VERSION:-1.4.2}
+ENV SBT_VERSION ${SBT_VERSION:-1.4.7}
 
 RUN apk --update add openjdk8 jq
 
 RUN gcloud components install kubectl
```
lkrystkowiak-ovoenergy, line 9: is kubectl still needed?

lkrystkowiak-ovoenergy (approved):
nice cleanup! 👍🏼 

might be good to update README as well

rac-gcp-deploy/executor/Dockerfile
```diff
@@ -2,27 +2,15 @@ ARG GCLOUD_SDK_VERSION
 FROM google/cloud-sdk:${GCLOUD_SDK_VERSION:-304.0.0-alpine}
 
 ARG SBT_VERSION
-ENV SBT_VERSION ${SBT_VERSION:-1.4.2}
+ENV SBT_VERSION ${SBT_VERSION:-1.4.7}
 
 RUN apk --update add openjdk8 jq
```
jensraaby, line 7: Is this where we can upgrade the Java version? I saw yesterday that parts of the latest FS2 release are compiled with Java 11, which failed a build, so now might be a time to upgrade

jensraaby: Update: they seem to have published a fix for that issue, but I wonder whether we have a plan for this? I've wanted to use some Java 9 APIs a couple of times and got burned once the code reached CircleCI

rac-gcp-deploy/executor/Dockerfile
rmorais: Good spot. CircleCi does not interact with k8s so this can go away.

rac-gcp-deploy/orb.yml
rmorais: Yes!!!

rac-gcp-deploy/executor/Dockerfile
rmorais: yes, this is the place.
I am not sure if compiling with java 11 will work for all of our services, but I believe it should be fine.
We can upgrade it and see how it goes or we can compile locally all of our services with java11 before hand to be on the safe side. :D

rmorais: Given the uncertainty, I will leave this for another PR

PR #235, opened 2021-02-18
Pin Clair Scanner to golang:1.15

chris-brindley:
This PR pins the base image in the clair scanner Dockerfile to `golang:1.15` instead of `golang:latest`. 1.16 was released a couple of days ago which seems to be breaking the build.


rmorais:
@ovotech/dex and @ovotech/production-eng can someone unblock this?

apjm (approved):

PR #237, opened 2021-02-19
Add chdir to plan out path

eversC:
Adds the `$chdir` variable to tf apply

If the user has chdir functionality enabled, the `plan.out` from the preceding tf plan would have been created in the `$chdir` directory (so currently all tf apply operations with chdir in the orb will be failing after the apply is done).

Users without chdir func enabled should be uneffected.

Tested the devorb on both paths (chdir enabled/disabled).

### DIFF
--- a/terraform/apply.sh
+++ b/terraform/apply.sh
@@ -22,12 +22,12 @@ function update_status() {
 }
 
 function apply() {
-    plan_path="plan.out"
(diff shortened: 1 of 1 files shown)


chris-brindley (approved):

eversC:
I'll rebase

I think we should look to merge this next week (not today) as it effects `tf apply` operations

PR #239, opened 2021-02-23
[GCP Rotate Keys] Decode base64 encoded keys if necessary

chris-brindley:
This PR updates the `rotate-key-redeploy-ecs-service` job in the `gcp-rotate-keys` orb to be able to decode the GCP service account key if it has been base64 encoded (for example, if it has been rotated by cloud key rotator).

### DIFF
--- a/gcp-rotate-keys/orb.yml
+++ b/gcp-rotate-keys/orb.yml
@@ -171,6 +171,12 @@ jobs:
       <<: *service-account
       <<: *ssm-path
     steps:
+      - run:
+          name: Prepare the GCP service account key
+          command: |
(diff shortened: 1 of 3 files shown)


eversC (approved):
lgtm

just out of interest with handling both base64-encoded and !base64-encoded, is that to handle if GCP switched (I believe they return base64-encoded currently)?

chris-brindley:
Yeah for user provided keys, they are generally just copy-pasted into the env variables as is so will often be non-b64 encoded, and then when for example CKR rotates them they become b64 encoded.

eversC:
great 👍

PR #242, opened 2021-02-26
Copy queue orb

olib963:
As noted in the README this is a copy of https://github.com/eddiewebb/circleci-queue, I would love to use this orb as part of a workflow to enforce that certain jobs cannot run concurrently, since at the moment this is on each developer to remember and forgetting can have quite disastrous effects. As such I've attempted to create a fork here.

I'm not sure that the CI setup is exactly the same in this repo as in the original repo so I might need help deploying a "dev" version of the orb to make sure it works before publishing anything. Let me know if there are any questions 😄

### DIFF
--- a/circleci-queue/README.md
+++ b/circleci-queue/README.md
@@ -5,34 +5,15 @@
 CircleCI Orb to limit workflow concurrency.
 
 Why? Some jobs (typically deployments) need to run sequentially and not parallel, but also run to completion. So CircleCI's native `auto-cancel` is not quite the right fit.
-See https://github.com/eddiewebb/circleci-challenge as an example using blue/green cloud foundry deployments.
+See https://github.com/ovotech/circleci-challenge as an example using blue/green cloud foundry deployments.
 
 ## Basic Usage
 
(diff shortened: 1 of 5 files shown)


whyman:
Could you not just use the upstream orb directly in your workflow?

olib963:
> Could you not just use the upstream orb directly in your workflow?

Yes I can, and in fact it works. I think the issue was that since it's externally controlled and this pipeline has full admin access to our production project, there's nothing stopping someone from randomly publishing a more damaging version of the orb to the open source Circle registry. E.g. send env to "me" silently. It was suggested that putting it here might be a good way of getting around that. Thoughts?

whyman:
It would be good to know if you can update a published revision of an orb?

Then pinning to a known good version could also alleviate this concern.

Personally I dont see how using this orb is much different from relying any other 3rd party library.

olib963:
I've found out that circle orbs are in fact immutable, so just going to use the third party one for now, I'll close this. If anyone feels strongly we should go this route instead let me know and I'll reopen

PR #251, opened 2021-04-01
Add a new orb: Pipeline utilities

chris-brindley:
This PR adds a new orb, tentatively called pipeline utilities. It is designed simply as a collection of commands that make life a bit easier when creating pipelines.

Currently there are two commands which should be well described in the README. Inspired by a [recent conversation on Slack](https://ovoenergy.slack.com/archives/CBFJV68US/p1617269997048100)

### DIFF
--- a/pipeline-utils/orb.yml
+++ b/pipeline-utils/orb.yml
@@ -39,4 +39,4 @@ commands:
           command: |
             include scripts/maybe-decode.sh
 
-            echo "export << parameters.var >>='"$(maybe_decode "$(printenv << parameters.var >>)")"'" >> "${BASH_ENV}"
+            maybe_decode "<< parameters.var >>"
--- a/pipeline-utils/orb.yml
+++ b/pipeline-utils/orb.yml
@@ -39,4 +39,4 @@ commands:
           command: |
             include scripts/maybe-decode.sh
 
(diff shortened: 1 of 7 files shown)


pipeline-utils/orb.yml
```diff
@@ -0,0 +1,42 @@
...
+      var:
+        description: The name of an environment variable to base64 decode
+        type: env_var_name
+    steps:
+      - run:
+          name: Base64 decode the environment variable
+          command: |
+            include scripts/maybe-decode.sh
+
+            echo "export << parameters.var >>='"$(maybe_decode "$(printenv << parameters.var >>)")"'" >> "${BASH_ENV}"
```
oarmstrong-ovo, line 42: Only nitpick is `configure-environment.sh` does this writing to `$BASH_ENV` and here you're doing it in the orb command instead. Might want both to follow the same pattern? Anyway it's a nitpick and this solves my problem so thanks! :bow:

oarmstrong-ovo (approved):

pipeline-utils/orb.yml
chris-brindley: Yep, makes sense to be consistent, will update

oarmstrong-ovo (approved):

PR #259, opened 2021-04-12
Remove sbt-dependency-graph

rmorais:
This plugin is now deprecated as it's been integrated directly into Sbt.
This plugin contains dependencies that are still hosted in Bintray. Given that the plugin is deprecated and Bintray is going to be terminated on 1st May, it's probable that the plugin is not going to be updated. If this is the case, it will cause our builds to fail.

Given that most of our services are using sbt > 1.4.x, I prefer to remove the injection of the plugin and update the services that might fail due to lack of it. 

Deprecation notice: https://github.com/sbt/sbt-dependency-graph


jensraaby (approved):
👍

PR #260, opened 2021-04-15
Terraform 0.15 compatibility with chdir

enicholson125:
The adding a terminal file path to the end of a terraform command method is no longer supported in terraform 0.15, only chdir should be used. However, there is a bug in our chdir implementation which puts empty quotes at the end of the command. This hopefully fixes it.

### DIFF
--- a/terraform/CHANGELOG.md
+++ b/terraform/CHANGELOG.md
@@ -3,6 +3,10 @@ All notable changes to the orb will be documented in this file.
 Orbs are immutable, some orb versions with no significant changes are
 not listed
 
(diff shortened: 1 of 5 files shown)


enicholson125:
This doesn't fix the issue :cry: Looking into it further

eversC:
Can we close this now we have `terraform-v2`?

enicholson125:
Yep, this is no longer relevant as terraform >= 0.14 is now supported only by the terraform-v2 orb, while the original orb supports <=0.14

PR #263, opened 2021-04-26
rac-gcp-deploy: Update location of SBT deb file - move away from bintray

jensraaby:


rmorais (changes requested):
SBT team maintains a url redirect service that should point to correct repository all the time. This is generally `repo.scala-sbt.org`.
To avoid issues in the future with repositories being moved, we should use that generic url.
We can fetch the debian binary using `https://repo.scala-sbt.org/scalasbt/debian/sbt-1.5.1.deb`

More details can be found https://eed3si9n.com/bintray-to-jfrog-artifactory-migration-status-and-sbt-1.5.1

jensraaby:
> SBT team maintains a url redirect service that should point to correct repository all the time. This is generally `repo.scala-sbt.org`.
> To avoid issues in the future with repositories being moved, we should use that generic url.
> We can fetch the debian binary using `https://repo.scala-sbt.org/scalasbt/debian/sbt-1.5.1.deb`
> 
> More details can be found https://eed3si9n.com/bintray-to-jfrog-artifactory-migration-status-and-sbt-1.5.1

Brilliant, I hadn't seen that blog! Will test this out on something, as a build was failing yesterday.

jensraaby:
Tested here: https://app.circleci.com/pipelines/github/ovotech/rac-gas-energy-consumption/839/workflows/3140a44f-63e2-4141-9cce-7597e587fa6f/jobs/1931

rmorais (approved):

PR #264, opened 2021-04-29
Tf 2

eversC:
okay this is a relatively large PR, so to summarise in case it helps:

- no functional changes to the existing `terraform` orb
- a change to `terraform` orb readme, with a note about how to upgrade to v2
- a new `terraform-v2` directory, containing new terraform orb code
- `terraform-v2` added to the publish orb matrix in circleci config
- the name of the branch has been added to the tag of the 'build' docker image that's pushed (the tag previously only has the current date in it, which could lead to conflict when multiple branches are being committed to on the same day). This should be fine for the official images built from master as this is done: `docker tag "ovotech/${ORB}:${BUILD_TAG}" "ovotech/${ORB}:${TAG}"` (meaning BUILD_TAG can be changed around, there's been no change to how `TAG` is calculated).


terraform-v2/orb.yml
```diff
@@ -0,0 +1,383 @@
+version: 2.1
+description: "Terraform orb. See https://github.com/ovotech/circleci-orbs/tree/master/terraform"
```
enicholson125, line 2:
```suggestion
description: "Terraform orb v2. See https://github.com/ovotech/circleci-orbs/tree/master/terraform-v2"
```

terraform-v2/orb_version.txt
```diff
@@ -0,0 +1 @@
+ovotech/terraform@2.0.0
```
enicholson125, line 1:
```suggestion
ovotech/terraform-v2@2.0.0
```

terraform-v2/test/aiven_provider/config.yml
```diff
@@ -0,0 +1,36 @@
+version: 2.1
```
enicholson125, line 1: I think this test isn't relevant to to the new orb?

terraform-v2/test/aiven_provider/test.sh
```diff
@@ -0,0 +1,32 @@
+#!/usr/bin/env bash
```
enicholson125, line 1: I think this test isn't relevant to the new orb

terraform-v2/test/fmt/config.yml
```diff
@@ -0,0 +1,57 @@
+version: 2.1
```
enicholson125, line 1: We should put in some terraform-v2 tests here

terraform-v2/v1-to-v2.md
```diff
@@ -0,0 +1,89 @@
...
+## How To Upgrade From V1 to V2
+
+You'll need to update the orb stanza in your .circleci/config.yml file:
+
+```yaml
+orbs:
+  terraform: ovotech/terraform-v2@1
+```
+
+This will use the latest 1.x.x version of the v2 orb, you can pin to a more
```
enicholson125, line 48: Minor thing, but are we going with 1.x.x for the first version of v2 orb? Would 2.x.x be more intuitive?

enicholson125 (changes requested):
Most of it is great but we should sort out the test directory

terraform-v2/README.md
```diff
@@ -0,0 +1,507 @@
...
+
+It contains:
+- google-cloud-sdk
+- aws-cli
+- Helm 3+ available as `helm` and `helm3` See [Using Helm 3](https://github.com/ovotech/circleci-orbs/tree/master/terraform-v2#Using-Helm-3)
+- stable "ovo" provider with ovo_kafka_user resource (`source = "terraform.ovotech.org.uk/pe/ovo"`)
+- stable "aiven-kafka-user" provider with aiven-kafka-users_user resource that enables auto-rotation of credentials  (`source = "terraform.ovotech.org.uk/pe/aiven-kafka-users"`)
+
+
+## Commands
```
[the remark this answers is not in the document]
eversC: can I just check how the doc in the orb registry would help in this case? There's no mention of the `aiven-kafka-users` tf provider on there?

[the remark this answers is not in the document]
enicholson125: Sorry, not a very well linked comment - I meant for the commands list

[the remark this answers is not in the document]
eversC: Ah gotcha, great idea, fixed in 0b21bdb

terraform-v2/test/fmt/config.yml
eversC: My plan is to try and get the tests already in terraform-v2 working, then at some point take a look at filling in any gaps, there's probably some cmds that aren't covered. They look really useful though.

terraform-v2/v1-to-v2.md
eversC: Yep I agree, fixed in dc0877d

enicholson125 (approved):

PR #268, opened 2021-05-14
Set tag for ASP Deploy Orb

puzzledbytheweb:


EduardoLaranjo (commented):
One thing that is making some confusion is the fact you have a default, so condition doesn't make much sense. What am I missing?

puzzledbytheweb:
> One thing that is making some confusion is the fact you have a default, so condition doesn't make much sense. What am I missing?

An empty string is a falsey value, I just wanted to have a bit more control there but I guess it's redundant since not having a value is also falsey.

EduardoLaranjo (approved):
LGTM 🚀

puzzledbytheweb:
Working in this PR - https://github.com/ovotech/circleci-orbs/pull/269

PR #270, opened 2021-05-20
Fix path of buildTag and releaseVersion

stephen-harris:
Values passed to a chart dependency need to be nested under the name of the chart


stephen-harris:
@rekaelek-ovo This (v1.9.0) would be a breaking change for your gitops deploy. But you no longer need to maintain your own templates and can use the helm chart https://github.com/ovotech/jaws-journey-helm-chart

Here's an example of how its implemented right now: https://github.com/ovotech/journey-meter-tariff-extractor/tree/master/manifests

ChrisFord-Ovo (approved):

PR #285, opened 2021-06-04
Use argocd orb

stephen-harris:


jaws-journey-deploy/orb.yml
```diff
@@ -625,6 +626,12 @@ jobs:
         default: "Journeys CircleCI Orb"
       ssh_key_fingerprint:
         type: string
+      argocd_url:
+        type: string
+        default: ""
```
ChrisFord-Ovo, line 631: i know this is here for backwards compatibility, but feels a bit dirty.  Do we guard anywhere for the empty string, or does the usage of it cause a clear error?

ChrisFord-Ovo (commented):
1 nitpick

PR #290, opened 2021-06-07
Prompt terraform 0.15 users to use the v2 orb

enicholson125:
The terraform fails with an unhelpful message currently if you try to use it with terraform 0.15+. This makes the error message more explicit and helpful.


eversC (approved):
looks good

PR #293, opened 2021-06-15
Feat/add publish job

ChrisFord-Ovo:
add job to orb so we can just publish images rather than having to build and publish - this is for cases of wanting to push the same image to multiple ecrs


jaws-journey-deploy/orb.yml
```diff
@@ -403,6 +403,44 @@ jobs:
...
+          aws-access-key-id: AWS_ACCESS_KEY_ID
+          aws-secret-access-key: AWS_SECRET_ACCESS_KEY
+          dockerfile: ./<< parameters.serviceName >>/Dockerfile
+          path: ./<< parameters.serviceName >>
+          create-repo: false
+          region: AWS_DEFAULT_REGION
+          repo: << parameters.serviceName >>
+          tag: "latest,$CIRCLE_SHA1"
+      - snyk/scan:
+          docker-image-name: $AWS_ACCOUNT_URL/<< parameters.serviceName >>:latest
```
stephen-harris, line 438: Could there be a race condition here if multiple jobs are running at once? Could we use `$CIRCLE_SHA1` instead?

stephen-harris (changes requested):

stephen-harris (approved):

PR #295, opened 2021-06-18
Test parallelism

eversC:
This adds args that are specific to apply stage

It follows the same model used for plan stage args; a `set_apply_args.sh` that adds orb params to an env var that's later used in the apply cmd

I tested the apply works in the dev orb (using `set -x` in `apply.sh` so I could see the resulting cmd)


terraform/CHANGELOG.md
```diff
@@ -3,6 +3,9 @@ All notable changes to the orb will be documented in this file.
 Orbs are immutable, some orb versions with no significant changes are
 not listed
 
+## ovotech/terraform@1.8.9
+## Added
```
enicholson125, line 7: Might be worth explicitly putting this as a bugfix, so that it's easier for people who come across the bug in older versions of the orb to find the fixed version?

enicholson125 (approved):
Awesome!

terraform/CHANGELOG.md
eversC: updated in fb9211b

PR #297, opened 2021-06-18
Update publish module script to prevent overwriting previous versions

stephen-harris:
Currently if a terraform module is published without bumping the module
version it will overwrite the existing version. That could cause issues
for users of that module if it includes breaking changes.

This checks if a version already exists before publishing.


eversC (approved):
looks good

PR #302, opened 2021-06-28
Comms team orb for publishing a service

d-fernandes:
As the title suggests, a new orb for the comms team to share CCI code across comms services

### DIFF
--- a/comms-publish-service/orb.yml
+++ b/comms-publish-service/orb.yml
@@ -1,0 +1,57 @@
+version: 2.1
+description: "Orb for deploying a service (adapted to the Comms team only)"
+
+executors:
+  docker-executor:
+    docker:
+      - image: circleci/openjdk:14.0.2-jdk-buster
+
+commands:
+  generate_sbt_cache_key:
+    description: "Generate sbt cache key"
+    steps:
+      - run: md5sum project/build.properties project/plugins.sbt build.sbt > .sbt_cache_key
+
+  restore_sbt_cache:
+    description: "Restore sbt cache"
+    steps:
+      - generate_sbt_cache_key
+      - restore_cache:
+          keys:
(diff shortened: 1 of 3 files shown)


comms-publish-service/orb.yml
```diff
@@ -0,0 +1,57 @@
...
+          docker_layer_caching: true
+      - checkout
+      - run: docker login --username << parameters.docker_user >> --password << parameters.docker_user_password >>
+      - attach_workspace:
+          at: .
+      - restore_sbt_cache
+      - run:
+          name: Publish docker image to ECR
+          command: |
+            sbt service/publish
```
joedarby, line 51: This will work for the Templates API where the code is in a subdirectory called `service`, but for our other projects the command is just `sbt publish`

comms-publish-service/orb.yml
```diff
@@ -0,0 +1,57 @@
...
+    description: "Restore sbt cache"
+    steps:
+      - generate_sbt_cache_key
+      - restore_cache:
+          keys:
+            - sbt-{{ checksum ".sbt_cache_key" }}
+            - sbt
+
+jobs:
+  publish_service:
```
joedarby, line 25: I like the concept of this orb. Are there other jobs we could add to it like build, unit test, integration test, etc?

joedarby (changes requested):
See comments inline

comms-publish-service/orb.yml
d-fernandes: Nice spot, will fix

comms-publish-service/orb.yml
d-fernandes: Definitely, I was planning to do that later when I know that this works

joedarby (approved):

mike-gregory-kaluza (approved):

PR #308, opened 2021-07-05
Comms Job for applying Terraform

d-fernandes:


comms-commands/orb.yml
```diff
@@ -1,6 +1,36 @@
...
+        description: "This can be either one of (prd, uat or lt)"
+        type: enum
+        enum: [ "prd", "uat", "lt" ]
+    executor: terraform/default
+    steps:
+      - checkout
+      - terraform/apply:
+          path: << parameters.path >>
+          workspace: main
+          backend_config_file: ./environments/<< parameters.environment >>/<< parameters.environment >>.backend
```
joedarby, line 23: This forces us to use this file structure in every service, but I think that's a good thing.

joedarby (approved):

comms-commands/orb.yml
l-catallo: not sure about this, ideally we should be free to split configs or change the structure when necessary

joedarby: Could have optional param then

l-catallo (approved):

PR #310, opened 2021-07-06
Default parameters do not support ref to other params (alternative)

d-fernandes:


comms-commands/orb.yml
```diff
@@ -15,25 +15,14 @@ jobs:
         description: "This can be either one of (prd, uat or lt)"
         type: enum
         enum: [ "prd", "uat", "lt" ]
```
l-catallo, line 17: I think it would be better to remove environment and keep the paths. It would be more context agnostic that way.

l-catallo (approved):

PR #316, opened 2021-07-09
Add Terraform 1.0.X compatible check

daogilvie:
'An execution plan....' does not appear in the output of Terraform plan/apply
for version 1.0.1 — this means that the simple examples of a plan-on-branch,
apply-on-main flow have high chances of never working:

The full plan output will be used for an exact string comparison, and it just
so happens that resource states can be refreshed in different orders. This means
that if comparing the two, they will often not match, despite the resulting
plan being completely identical in effect.

Adding 'Terraform will perform the following actions' will clip out just the lines
with the actions!

What do I need to do to get this in to a new version of the executor please?
Happy to get stuck in 😄 

See https://app.circleci.com/pipelines/github/ovotech/foyer-infra/31/workflows/98b82903-942c-4b53-afd0-8d1c8e3f7c84/jobs/36 for an example of this happening.

PR #319, opened 2021-07-19
Add Terraform 1.0.X compatible check to compact_plan

enicholson125:
This is from https://github.com/ovotech/circleci-orbs/pull/316. I've pushed the branch to the main repo so that the tests can run and to build a dev version of the new executor for manual testing.

'An execution plan....' does not appear in the output of Terraform plan/apply
for version 1.0.1 — this means that the simple examples of a plan-on-branch,
apply-on-main flow have high chances of never working:

The full plan output will be used for an exact string comparison, and it just
so happens that resource states can be refreshed in different orders. This means
that if comparing the two, they will often not match, despite the resulting
plan being completely identical in effect.

Adding 'Terraform will perform the following actions' will clip out just the lines
with the actions!

What do I need to do to get this in to a new version of the executor please?
Happy to get stuck in smile I've added a happy-path test for the use case that
triggered my involvement here.

See https://app.circleci.com/pipelines/github/ovotech/foyer-infra/31/workflows/98b82903-942c-4b53-afd0-8d1c8e3f7c84/jobs/36 for an example of this happening.

### DIFF
--- a/terraform/executor/compact_plan.py
+++ b/terraform/executor/compact_plan.py
@@ -10,6 +10,7 @@ def compact_plan(input):
     for line in input:
 
         if not plan and (
+            line.startswith('Terraform used the selected providers to generate') or
             line.startswith('An execution plan has been generated and is shown below') or
             line.startswith('No changes') or
             line.startswith('Error')
--- a/terraform/executor/test_compact_plan.py
+++ b/terraform/executor/test_compact_plan.py
@@ -256,3 +256,62 @@ This should protect against the output changing again."""
     output = '\n'.join(compact_plan(input.splitlines()))
     assert output == expected_output
 
+def test_plan_1_0():
+    input_one = """
+STATE_REFRESH_1
+STATE_REFRESH_2
+
+Terraform used the selected providers to generate the following execution
+plan. Resource actions are indicated with the following symbols:
+  + create
+
+Terraform will perform the following actions:
+
+
+# random_string.my_string will be created
++ random_string.my_string
+      + id:          <computed>
+      + length:      "11"
+      + lower:       "true"
+      + min_lower:   "0"
+      + min_numeric: "0"
+      + min_special: "0"
+      + min_upper:   "0"
+      + number:      "true"
+      + result:      <computed>
+      + special:     "true"
+      + upper:       "true"
+
+Plan: 1 to add, 0 to change, 0 to destroy."""
+
+    input_two = """
+STATE_REFRESH_2
+STATE_REFRESH_1
+
+Terraform used the selected providers to generate the following execution
+plan. Resource actions are indicated with the following symbols:
+  + create
+
+Terraform will perform the following actions:
+
+
+# random_string.my_string will be created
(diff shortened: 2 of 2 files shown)


enicholson125:
> Hi there, any chance this might be mergeable?

Yes, I'll review this today. Thanks for the prompt and apologies that it's taken so long.

enicholson125:
Approved on the [original PR](https://github.com/ovotech/circleci-orbs/pull/316/files)

eversC:
I just had to get involved in the fun here and on #316 :O

For what it's worth, I had a load of my own issues with the orb, involving a lot of SSHing into the orb/container in CircleCI (I created [a testing repo](https://github.com/ovotech/circleci-orbs-testing) which I hope turns into an easy way to e2e test dev orbs in future). I think we could also improve unit tests in this repo to help contributors.

Those issues made testing pretty hard, so apologies if there's something I've removed that shouldn't have been.

One of the things removed was the use of `difflib`:

```python
diff = list(
    difflib.ndiff(
        generated_plan.splitlines(keepends=True), plan_from_pr.splitlines(keepends=True)
    )
)
```

was this to get round a specific issue? In theory the simple comparison of `generated_plan == plan_from_pr` should be fine as far as I'm aware. I'm happy to put it back in and give the dev orb another test.

I can see there was a commit to add some debig printlines into `github.py` (I actually added those back in in some parts of my debug trail). Unfortunately that breaks the diff as the print statements are outputted into the diff file output, causing the diff check in `cmp.py` to fail (you may have alluded to this in #316).

The orb is supposed to update comments it's previously added to the PR in question, so that when merge to main branch happens there's only a single comment for it to do the diff check on. I think there's a flake that's been around for a while, though, which can lead to multiple PR comments, and possibly a failure in the diff check.

enicholson125 (commented):
Thanks for making these changes, Chris! They look sensible and your testing in circleci-orbs-testing looks good to me. I raised the PR, so can't formally approve but definitely informally approve.

enicholson125:
A version bump would be good :)

InBrewJ:
Hi @ovotech/production-eng, my team has been testing out the `ovotech/terraform:terraform-1.0.X-compact-plan_31-08-2021` executor and the `ovotech/terraform-v2@dev:terraform-1.0.X-compact-plan` orb for some weeks now.

I've run our terraform CI pipeline a few times today. I tested out things like this via TF:
- Adding (and then removing) a secret to both `nonprod` and `prod`
- Adding an S3 bucket on `nonprod` but **not** on `prod`
- Removing that same S3 bucket from `nonprod`

Everything worked like a dream and I didn't see any cryptic parsing errors, just the errors I expected from the TF runtime. This is a godsend; we haven't had a functioning CI pipeline for a while now.

If you're willing, can we get some approvals on this 🙏🏻? If so, OVO France and OVO Spain would be in your debt...

eversC:
@daogilvie + @InBrewJ  now this has been merged/published, please let me/ProdEng know if you see any weirdness

PR #320, opened 2021-07-20
comms commands - change tf default file paths to match preferred stru…

joedarby:
…cture

### DIFF
--- a/comms-commands/orb.yml
+++ b/comms-commands/orb.yml
@@ -22,7 +22,7 @@ jobs:
     steps:
       - terraform_plan:
           path: << parameters.path >>
-          backend_config_file: ./backends/<< parameters.environment >>.backend
+          backend_config_file: ../backends/<< parameters.environment >>.backend
           var_file: ./variables/<< parameters.environment >>.tfvars
           image_version: << parameters.image_version >>
 
@@ -47,7 +47,7 @@ jobs:
             - terraform_apply:
                 path: << parameters.path >>
                 run_shipit: true
-                backend_config_file: ./backends/<< parameters.environment >>.backend
+                backend_config_file: ../backends/<< parameters.environment >>.backend
                 var_file: ./variables/<< parameters.environment >>.tfvars
                 image_version: << parameters.image_version >>
       - when:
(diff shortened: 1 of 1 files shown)


comms-commands/orb.yml
```diff
@@ -19,8 +19,8 @@ jobs:
     steps:
       - terraform_plan:
           path: << parameters.path >>
-          backend_config_file: ./environments/<< parameters.environment >>/<< parameters.environment >>.backend
-          var_file: ./environments/<< parameters.environment >>/<< parameters.environment >>.tfvars
+          backend_config_file: ./backends/<< parameters.environment >>.backend
```
l-catallo, line 22: wouldn't this still force you to duplicate the backend files for each module?

joedarby: Good point. Ok I think `backend_config_file` and `var_file` are both relative to `path`. So how about this:

`backend_config_file: ../terraform-backends/<< parameters.environment >>.backend`

Then in a single module project we just have:

```
terraform
├── variables
│   ├── prd.tfvars
│   └── uat.tfvars
└── main.tf
terraform-backends
├── prd.backend
└── uat.backend
```

And in a multiple module project we have:

```


terraform
├── module-A
│   ├── variables
│   │   ├── prd.tfvars
│   │   └── uat.tfvars
│   └── main.tf
├── module-B
│   ├── variables
│   │   ├── prd.tfvars
│   │   └── uat.tfvars
│   └── main.tf
└── terraform-backends
    ├── prd.backend
    └── uat.backend

```

l-catallo: Yeah I'm ok with that. I'd try to keep it simple and only use structure n. 2.
We can treat single module projects as multi-module with just 1 module.

I don't think that would make a difference for this PR anyway, just mentioning.

l-catallo (approved):

comms-commands/orb.yml
```diff
@@ -58,7 +58,7 @@ jobs:
             - terraform_apply:
                 path: << parameters.path >>
                 run_shipit: false
-                backend_config_file: ./environments/<< parameters.environment >>/<< parameters.environment >>.backend
+                backend_config_file: ../environments/<< parameters.environment >>/<< parameters.environment >>.backend
                 var_file: ./environments/<< parameters.environment >>/<< parameters.environment >>.tfvars
```
l-catallo, line 62: I think these need to be updated to the new paths as well

joedarby: Oh I think I messed up my rebase :(
Will fix

l-catallo (approved):

PR #321, opened 2021-07-20
Pass image version to tf plan and apply. On publish tag image with pa…

joedarby:
…ssed in value. Delete unnecessary tag command.


rcdmrl (approved):
Makes sense

PR #323, opened 2021-07-21
Implement path filtering on circle ci

joedarby:


eversC (approved):
lgtm

we can always review again if we need to establish some 'required checks' on branch protection.

PR #325, opened 2021-07-22
Bump aws cli version.

sundeepkovo:

PR #326, opened 2021-07-22
Bump aws cli version.

eversC:
commits from #325

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@ any changes to the orb.
 
  - [ovotech/argocd](argocd) - Interact with an ArgoCD API.
  - [ovotech/aws-get-parameters@1](aws-get-parameters) - Get parameters from AWS Parameter Store.
- - [ovotech/aws-rotate-keys@1](aws-rotate-keys) - Rotate AWS access keys and update corresponding CircleCI environment variables.
+ - [ovotech/aws-rotate-keys@2](aws-rotate-keys) - Rotate AWS access keys and update corresponding CircleCI environment variables.
  - [ovotech/clair-scanner@1](clair-scanner) - Scan Docker images for vulnerabilities.
  - [ovotech/rac-gcp-deply@1](rac-gcp-deploy) - Deploy Scala services to a Kubernetes cluster running in Google Cloud.
  - [ovotech/set-current-environment@1](set-current-environment) - Set current environment (master branch, tag or other) to environmental variable.
--- a/aws-rotate-keys/CHANGELOG.md
+++ b/aws-rotate-keys/CHANGELOG.md
@@ -1,0 +1,10 @@
+# Changelog
+All notable changes to the orb will be documented in this file.
+Orbs are immutable, some orb versions with no significant changes are
+not listed
+
+## ovotech/aws-rotate-keys@2.0.0
+ - Updated the orb to use the aws-cli orb v2.
+
+## ovotech/aws-rotate-keys@1.x.x
+- Start of the changelog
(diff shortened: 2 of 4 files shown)


mwz:
@sundeepkovo the dev version has now been published to the orb registry so feel free to test your changes just in case the version bump causes any issues.

I don't think this needs a major version bump since the aws-cli/install job was used to install aws-cli 2, so technically there shouldn't be be any impact downstream?

eversC:
hey @mwz thanks for the info

do you have any idea where the version of the orb comes from? Usually we have an `orb_version.txt` file in the orb dir in this repo, but I notice there isn't one for the `aws-rotate-keys` orb but it's been published with a version of `1.0.226` https://circleci.com/developer/orbs/orb/ovotech/aws-rotate-keys

in terms of the impact downstream, I guess the cli version bump could cause the rotation to break (?) but so long as there's a succesful test with the devorb then it's good to go

mwz:
> do you have any idea where the version of the orb comes from?

it never had the orb_version.txt file afaik so it was always published with the version bumped automatically - we should definitely add it


> in terms of the impact downstream, I guess the cli version bump could cause the rotation to break (?)

my suspicion is that it won't because the `aws-cli/install` job is run with the version 2 of aws-cli ([here](https://github.com/ovotech/circleci-orbs/blob/master/aws-rotate-keys/orb.yml#L33-L35)) so I don't think there will be any impact, but we could bump the version just to be safe.

eversC:
> my suspicion is that it won't because the aws-cli/install job is run with the version 2 of aws-cli (here) so I don't think there will be any impact, but we could bump the version just to be safe.

what if there's been changes to the `aws iam` cmds that your orb uses? (there probably hasn't, so I expect this is a moot point)

mwz:
I think we're ok based on https://docs.aws.amazon.com/cli/latest/userguide/cliv2-migration.html, but I wouldn't mind bumping the major version just to be safe.

sundeepkovo:
@eversC / @mwz I think line 33 should also be `- aws-cli/setup:` according to the [doc](https://circleci.com/developer/orbs/orb/circleci/aws-cli) it no longer is install. I don't have permissions to push to the branch so I cannot push.

mwz:
> according to the [doc](https://circleci.com/developer/orbs/orb/circleci/aws-cli) it no longer is install. I don't have permissions to push to the branch so I cannot push.

https://circleci.com/developer/orbs/orb/circleci/aws-cli?version=2.0.3#commands-install

`aws-cli/install` still exists in the v2 of the orb - it installs the cli if it isn't already installed without configuring the credentials file.

sundeepkovo:
Hi @mwz, do you think we can merge the change? Works well when I tested it out.

mwz:
@sundeepkovo if you could add an `orb_version.txt` file with `ovotech/aws-rotate-keys@2.0.0` in case of any aws cli incompatibilities and update version references in the readme to `@2` that would be great.

I think it would also be great to add a little changelog section to the bottom of the readme and mention that we've upgraded the aws-cli orb to v2 in the latest version.

Would you be able to make the above changes?

sundeepkovo:
@mwz I can do that, just need permissions to push.

mwz:
> @mwz I can do that, just need permissions to push.

Thank you, can you please request permissions in #developer-exp on Slack?

eversC:
@sundeepkovo sorry only just caught up with all this, I've given you `write` access to the repo now 👍

sundeepkovo:
Addressed changes, should be ready for review.

mwz (approved):

sundeepkovo:
@ovotech/production-eng / @ovotech/dex can I get a review? Little blocked on this.

SteveRothwell (approved):

PR #329, opened 2021-07-27
Add a wait_for_ecs_deployment command

l-catallo:
Allows for waiting the completion of an ECS deployment without restarting the
service.


rcdmrl (approved):
Makes sense

d-fernandes (approved):
Makes sense

PR #332, opened 2021-08-06
Ignore terraform changes to latest_restorable_time

oarmstrong-ovo:
Terraform 0.15.4 starts to report upon the drift changes that have been
detected rather than silently updating its state. One of the attributes
that is now reported includes the latest_restorable_time attribute for RDS
databases.

These drift changes are reported as part of the Terraform plan. Thus,
these changes are also included in the comparison when checking for
approval.

For AWS computed attributes this causes a problem. PITR and database
backups can constantly modify the latest_restorable_time attribute,
meaning that almost every single plan returns a more recent time.
Therefore it's almost impossible for this attribute to ever match what
was in the approved plan, as even if only a few minutes have passed
since the plan was approved the attribute has likely updated again.

We can filter out any lines in the plan that refer to the
latest_restorable_time attribute so that the value mismatch between the
approved plan and the apply-time plan isn't relevant.

### DIFF
--- a/terraform-v2/cmp.py
+++ b/terraform-v2/cmp.py
@@ -1,9 +1,20 @@
 #!/usr/bin/env python3
 
 import sys
+import re
 
 with open(sys.argv[1], encoding='utf-8') as f:
     first = f.read()
 with open(sys.argv[2], encoding='utf-8') as f:
     second = f.read()
-exit(0 if first.strip() == second.strip() else 1)
+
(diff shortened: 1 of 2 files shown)


eversC (approved):
lgtm

sundeepkovo (approved):

sundeepkovo:
Any update one when we can see this merged?

PR #335, opened 2021-08-24
Add init_reconfigure yes/no parameter

daogilvie:
~~This will enable use-cases like the Foyer project, where one infra
definition is applied to multiple projects within the same session.
See https://stackoverflow.com/a/59781224/1812006 as source for idea.~~

Turns out this was user error. Thanks folks!


daogilvie:
Closing as not needed

PR #342, opened 2021-10-04
RNC-2199: add snyk container scanning following push to registry

nigelhp:


lkrystkowiak-ovoenergy (approved):
lgtm 👍🏼

rmorais:
this new command should be used before or after the image is uploaded to the register?

nigelhp:
> this new command should be used before or after the image is uploaded to the register?

The current implementation is scanning **after**, which is inline with the reference implementation:
https://github.com/ovotech/circleci-orbs/blob/ad79525f8ea856d9ebed6a0ebd3ad4e97d1f3880/jaws-journey-deploy/orb.yml#L417

The image name would have to the full registry-qualified name in the case where you are instructing snyk to setup regular scanning (i.e. `monitor-on-build: true`) - note that we do not currently have this flag set.  I had thought that meant that the image needed to already be in the registry.  However, as we are tagging the image locally with the fully qualified name prior to pushing, I assume that the scan is able to use the local image.  Assuming this is the case, I agree that it would seem preferable to scan prior to the push in order to avoid polluting the registry.

rmorais:
If the main idea is to increase security, then we should only push images after they are scanned and are "OK". 
IMHO, this would be the right way to go, specially if we wanted to fail the build, which we are not going to do for now.

Given that we aren't failing the build, it doesn't matter much if we do it before or after, but if we can scan it locally, then it would make our lives easier in the future.

Other thing to take into account, is that once the images are published to the registry, flux will update the relevant deployments with the new images.

nigelhp:
Using `rac-rui-playground` as a test mule, testing output can be found here:

- [release task](https://app.circleci.com/pipelines/github/ovotech/rac-rui-playground/105/workflows/91951292-bd25-4c79-b043-55788734cc30/jobs/292)
- [push_docker_image task](https://app.circleci.com/pipelines/github/ovotech/rac-rui-playground/102/workflows/79b9968c-123c-4d21-afc1-af7f053fdcc6/jobs/289)

PR #347, opened 2021-10-19
[aws-rotate-keys] Default to installing the latest AWS CLI v2 if the binary is not present on the PATH

mwz:
This fixes the issue with the version `'2'` that was specified previously which made the job fail if the default executor was used since the install command requires a full version to be specified. I've made it use the latest v2 version by omitting the version parameter.


mwz:
@ovotech/dex any chance I could get a review on this?

bnjns (approved):
LGTM

PR #356, opened 2021-11-04
Add upload jacoco artifacts option jaws

fionamatters:


robturnerovo (approved):
👍

jaws-journey-deploy/orb.yml
```diff
@@ -537,8 +549,12 @@ jobs:
...
                 path: ~/test-results/junit
                 destination: ~/reports/diagrams
-            - store_artifacts:
-                path: ~/reports/jacoco
+            - when:
+                condition: << parameters.store_jacoco_artifacts >>
+                steps:
+                  - store_artifacts:
+                      name: store jacoco artifacts
+                      path: ~/reports/jacoco
```
ilwallace, line 557: Is this the right path? everything else is using ./reports/jacoco - obviously this would be an existing bug.

PR #365, opened 2021-12-20
Add KMI image push job to RaC orb

enicholson125:
This adds a new job to the R&C orb, to push an image to the centralised image registry where it can be pulled in by KMI. The AWS_ECR_ACCOUNT_URL and AWS_REGION environment variables have already been added to the main R&C context but the AWS image pushing credentials will need to be added per service, as they are different for each namespace.

I have not added this new job to the README as the image pushing jobs don't seem to be in there and I wanted to double check if this is a deliberate choice, given the orb is public?

I'm also not sure how to increase the version for this orb, as I can't see an orb_version.txt file?

### DIFF
--- a/rac-gcp-deploy/orb.yml
+++ b/rac-gcp-deploy/orb.yml
@@ -16,6 +16,7 @@ aliases:
 
 orbs:
   snyk: snyk/snyk@0.0.13
+  aws-ecr: circleci/aws-ecr@7.2.0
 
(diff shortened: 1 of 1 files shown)


enicholson125:
Closing in favour of the identical https://github.com/ovotech/circleci-orbs/pull/366, as orb caching is getting in the way of testing.

PR #370, opened 2021-12-23
Add Helm CI orb

enicholson125:
Teams moving to KMI are being encouraged to use helm in their gitops repositories. I'm finding there's a lot of duplication in the CI for those gitops repositories and there doesn't seem to be a good open source circleCI orb for Helm CI, so creating one.

This is quite focused on the KMI use-case - e.g. it assumes you will have values files associated with the chart and no values being passed in directly in the helm templating step. I think this is acceptable for a first version of the orb and we can add parameterisation for other use cases in future.

https://app.circleci.com/pipelines/github/ovotech/energy-contracts-kmi-gitops/104/workflows/f0ca0afe-3d3e-4c54-9e6b-2faca77a8e7b/jobs/187 is an example of the orb running.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -20,6 +20,7 @@ workflows:
             clojure/.* run-clojure true
             comms-commands/.* run-comms-commands true
             gcp-rotate-keys/.* run-gcp-rotate-keys true
+            helm-ci/.* run-helm-ci true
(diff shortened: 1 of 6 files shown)


stephencoe (approved):

gavinmcnair (approved):
LGTM. Like the use of helm template, lint and kubeval 💯

PR #375, opened 2022-02-04
Fix Helm example

eversC:


enicholson125 (commented):
Looks good to me - just a couple of suggestions to not use a bucket name at all in the README

eversC:
> Looks good to me - just a couple of suggestions to not use a bucket name at all in the README

@enicholson125 are there any more changes required here? I couldn't quite tell whether this comment was a suggestion in itself, further the to the GitHub suggestions that were made..

I guess if there's no bucket name used (with no `helm repo add` cmd either), it renders the example incomplete and could confuse users

enicholson125 (approved):
The bucket names in the example were all I meant - it looks good to me :)

PR #381, opened 2022-02-17
specify machine image in jaws deploy orb

Adam-McDevitt:


jaws-journey-deploy/orb.yml
```diff
@@ -627,6 +627,7 @@ jobs:
     resource_class: << parameters.resourceClass >>
     machine:
       docker_layer_caching: false
+      image: ubuntu-2004:202201-02
```
ChrisFord-Ovo, line 630: what is the purpose of this change?

ChrisFord-Ovo: would it not be better to pass this in rather than hardcoding? i.e. << parameters.image >>

Adam-McDevitt: yeh probably makes sense to just specify it once. I'll try something like << parameters.image >>

Adam-McDevitt: @ChrisFord-Ovo pushed up some changes to stick this in a parameter

ChrisFord-Ovo: cool - yeah also means consumers of the orb could change the image if they wanted - although not sure why

ChrisFord-Ovo (approved):
lgtm

ChrisFord-Ovo (approved):
wont you need to bump the version of the orb actually

PR #384, opened 2022-02-24
RNC-2519: R&C orb for KMI deployments

nigelhp:


enicholson125 (changes requested):
Would you please add an orb_version.txt file to help with versioning? [This is an example](https://github.com/ovotech/circleci-orbs/blob/master/oot-deploy/orb_version.txt) of the expected format. Otherwise this looks good to me :)

rmorais:
added the version file

enicholson125 (approved):

PR #391, opened 2022-03-04
Update image for machine executor

rmorais:
Older images will soon be deprecated by CircleCi
https://circleci.com/blog/ubuntu-14-16-image-deprecation/

### DIFF
--- a/rac-gcp-deploy/orb.yml
+++ b/rac-gcp-deploy/orb.yml
@@ -20,7 +20,8 @@ orbs:
 
 executors:
   machine:
-    machine: true
+    machine:
+      image: ubuntu-2004:202201-02
   docker:
     docker:
       - image: ovotech/rac-gcp-deploy:latest


rac-gcp-deploy/orb.yml
```diff
@@ -20,7 +20,8 @@ orbs:
 
 executors:
   machine:
-    machine: true
+    machine:
```
jensraaby, line 23: Is this basically redefining a built in called "machine"? Just trying to work out how to interpret this nesting

rmorais: `machine` is the default CircleCi executor which uses a default image. That default image is deprecated and we need to specify an up to date image.

rmorais: I might misinterpreted your question. 
We are just naming an executor `machine` which _uses_ the default  `machine` executor. We do this so we are able to reference to it outside of the orb

rmorais: @jens can you approve this or is there anything you would like to change?

jensraaby (approved):

rac-gcp-deploy/orb.yml
jensraaby: Nope that's fine I just missed your replies 🙈

PR #404, opened 2022-04-04
Add orb to authenticate via CircleCI OIDC

stephen-harris:


.github/CODEOWNERS
```diff
@@ -1,21 +1,22 @@
...
+/oot-eks/                        @ovotech/orion-ops-tooling
+/clojure/                        @ovotech/orion-ops-tooling
+/ssh-proxy/                      @ovotech/orion-security-engineering
+/jaws-journey-deploy             @ovotech/jaws-library-maintainers
+/oot-deploy/                     @ovotech/orion-ops-tooling
+/sbt-deploy/                     @ovotech/smart-bookings
+/asp-deploy/                     @ovotech/asp-team
+/argocd/                         @ovotech/metering-industry-abstraction-cluster-admins
+/comms-commands/                 @ovotech/comms
+/ipa-deploy/                     @ovotech/ipa
```
enicholson125, line 22: Not sure why Github's complaining about these - that team definitely exists (although it also hasn't changed in this PR so out of scope)

aws-configure-credentials-oidc/orb.yml
```diff
@@ -0,0 +1,22 @@
...
+        description: The ARN of the IAM Role to assume (e.g. arn:aws:iam::999999999999:role/CircleCI).
+        type: string
+      profile-name:
+        default: default
+        description: Profile name to be configured.
+        type: string
+    steps:
+      - aws-cli/install
+      - run: |
+          STS=($(aws sts assume-role-with-web-identity --role-arn << parameters.role-arn >> --role-session-name "CircleCI-${CIRCLE_PROJECT_REPONAME}-${CIRCLE_BUILD_NUM}" --web-identity-token $CIRCLE_OIDC_TOKEN --duration-seconds 3600 --query 'Credentials.[AccessKeyId,SecretAccessKey,SessionToken]' --output text))
```
enicholson125, line 19: Might be worth making the duration of the role assumption configurable? If people are using this to apply Terraform, some resource creation can take a while. (3600 seconds seems a good default though)

aws-configure-credentials-oidc/orb_version.txt
```diff
@@ -0,0 +1 @@
+ovotech/aws-configure-credentials-oidc@0.1.0
```
enicholson125, line 1: I believe circleCI overwrites the version to 1.0.0 on the first publish, if it's 0.x.x (I couldn't find any docs stating that, but it looked like that happened to my last new orb)

enicholson125 (approved):
Very excited about this orb!

.github/CODEOWNERS
stephen-harris: Yeah, I noticed that too. Seems to be complaining about some already: https://github.com/ovotech/circleci-orbs/blob/master/.github/CODEOWNERS

aws-configure-credentials-oidc/orb.yml
stephen-harris: Good point. One hour should be sufficient. It think for initial deployments (where there are changes required to route53-as-code it could fail and need to be re-run). Will follow up with a PR to make this configurable.

aws-configure-credentials-oidc/orb_version.txt
stephen-harris: Thanks. Updated to 1.0.0 to be on the safe side

PR #416, opened 2022-04-21
Jaws snyk scan threshold parameter

rekaelek-ovo:
With the security dashboard flagging medium vulnerabilities, we need the ability to flag them during builds


keirlawson (approved):
:heart_eyes:

PR #427, opened 2022-05-10
Update to using the Jukie repo for tfswitch in terraform docker container

kizzie:
There is a current issue with the original tfswitch as hashicorp changed
their website and so the bit of tfswitch which scrapes that info is
failing. This is a temporary fix until the PR is merged on tfswitch.

github issue: https://github.com/warrensbox/terraform-switcher/pull/225
differences between versions: https://github.com/warrensbox/terraform-switcher/compare/master...Jukie:fix-release-downloads

### DIFF
--- a/terraform/executor/Dockerfile
+++ b/terraform/executor/Dockerfile
@@ -3,7 +3,7 @@
 FROM debian:buster-slim
 
 ARG TFMASK_VERSION="0.7.0"
-ARG TFSWITCH_VERSION=0.11.1071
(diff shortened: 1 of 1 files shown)


kizzie:
Alrighty - so this may not be needed by tomorrow if the tfswitch people wake up and merge their PR - at which point we just need to update to the latest version. But opening this PR  so that we can merge the change and people can run their terraform still if that is going to take a while

kizzie:
Closing this to rename the branch so docker stops being unhappy :)

PR #428, opened 2022-05-10
Update to using the Jukie repo for tfswitch in terraform docker container

kizzie:
There is a current issue with the original tfswitch as hashicorp changed
their website and so the bit of tfswitch which scrapes that info is
failing. This is a temporary fix until the PR is merged on tfswitch.

github issue: https://github.com/warrensbox/terraform-switcher/pull/225
differences between versions: https://github.com/warrensbox/terraform-switcher/compare/master...Jukie:fix-release-downloads

---

Alrighty - so this may not be needed by tomorrow if the tfswitch people wake up and merge their PR - at which point we just need to update to the latest version. But opening this PR so that we can merge the change and people can run their terraform still if that is going to take a while

### DIFF
--- a/terraform/executor/Dockerfile
+++ b/terraform/executor/Dockerfile
@@ -3,7 +3,7 @@
 FROM debian:buster-slim
 
 ARG TFMASK_VERSION="0.7.0"
-ARG TFSWITCH_VERSION=0.11.1071
+ARG TFSWITCH_VERSION=0.13.12345
 ARG GCLOUD_VERSION=329.0.0
 ARG AWSCLI_VERSION=1.19.16
 ARG HELM3_VERSION=3.5.3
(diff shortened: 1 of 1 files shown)


toniokaluza:
Please merge this, the original issue is still ongoing, so it may well be another day before it's merged

eversC (approved):
Seeing as that terraform-switcher clone has been authored by someone outside the org, it'd be great to switch back to the official release as soon as possible (just stating the obvious here)

eversC:
Does tfswitch in the legacy `terraform` orb work in a slightly different way, i.e. they have the terraform binaries baked in, so is it possible tfswitch isn't failing in those executors? (I'm totally unsure and can't really test for another hour or so..)

kizzie:
Oh 100% my plan is to switch this back to the official one when they fix it (or we make our own fork and keep that up to date)

I'll merge this one for now to (hopefully) fix the terraform-v2 side and open another PR to do the original orb as well

eversC:
> Does tfswitch in the legacy `terraform` orb work in a slightly different way, i.e. they have the terraform binaries baked in, so is it possible tfswitch isn't failing in those executors? (I'm totally unsure and can't really test for another hour or so..)

ignore my comment here, `tfswitch` is actually used in the legacy Dockerfiles in the baking in, so the docker builds would probably fail

kizzie:
Yep the v1 things are failing as well so I've got the second PR just merged (#430) to fix that. Just need to remember to switch it all back to the official one when the issue is resolved

PR #454, opened 2022-06-20
increase version of tf module

ChrisFord-Ovo:
Update TF Orb version to fix tfswitch issue - https://ovoenergy.slack.com/archives/C7D5V1JMS/p1655713296162659


samcooper720x (approved):
I think the publish job is still running, hence failing step, but LGTM 👍

stephen-harris (approved):

PR #472, opened 2022-07-22
Print out information with Clojure orb that allows for easier diagnos…

kelveden:
…ing of snyk scanning problems.


retrojetpacks (approved):
looks great!

PR #475, opened 2022-07-27
No need to install mvn in the Clojure orb- it's already installed.

kelveden:


retrojetpacks (approved):
Looks great, I turst you know what you're up to. Can't see any syntax errors:)

PR #476, opened 2022-07-27
Force a snyk monitor whenever building Clojure projects

kelveden:
This ensures that any violations that are picked up by the scanning step are available to suppress in the dashboard - currently the `monitor-on-build` option only kicks in if the scan itself passes.


retrojetpacks (approved):
looks good to me:)

PR #477, opened 2022-07-27
Avoid hitting snyk monitor twice in the Clojure orb

kelveden:
The `snyk/scan` implicitly will run a `monitor` if the main command succeeds. In our case, the command _is_ `monitor`, so it runs twice.

We could just disable the `monitor-on-build` but instead let's use an innocuous command `version` and let the snyk orb work out how to run a monitor.


retrojetpacks (approved):
Ok!

PR #478, opened 2022-08-03
[RNC-2769] Add optional prod approval step to release

jensraaby:


tomp4l (approved):
Looks good, what's this orb used for though?

tomp4l:
Spoke to soon, it's not valid schema :cry:

jensraaby:
> Spoke to soon, it's not valid schema 😢

You're too quick! I didn't realise it would add reviewers to the PR straight away.

It looks like I can't add a conditional "job" to another job, so I'm back to square one.
The "release" job was the original deploy workflow for everything until incident-gate. I noticed with the neo facade that the template was using it, so I was hoping rather than wiring together a whole new job I'd be able to just add in the approval bit as optional (in case something else is using it and we don't want to add manual approval).

tomp4l:
Makes sense, shame we can't do it like this, it looked sensible

PR #481, opened 2022-08-10
Fix authorization bearer for existing terraform version

eversC:
from #479 fork

### DIFF
--- a/terraform-v2/CHANGELOG.md
+++ b/terraform-v2/CHANGELOG.md
@@ -3,6 +3,9 @@ All notable changes to the orb will be documented in this file.
 Orbs are immutable, some orb versions with no significant changes are
 not listed
 
+## ovotech/terraform-v2@2.4.17
+- Added authentication to the API call for `/versions`
+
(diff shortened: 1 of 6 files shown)


eversC:
did the `/versions` endpoint get made private at some point?

I think we may need to bump the orb version here too

Does the same change need to be made to the `terraform-v2` orb?

drewhemm:
Yes, the same change needs to be made to `terraform-v2` - apolologies, as the PR should have included both. I have pushed a second commit to my fork

eversC:
> Yes, the same change needs to be made to `terraform-v2` - apolologies, as the PR should have included both. I have pushed a second commit to my fork

looks good

I think the orb versions (+changelogs) need updating, as that way the changes will get automatically released when we merge.

drewhemm:
I've updated my fork with the changes to the orb versions and changelogs:

https://github.com/drewhemm/circleci-orbs/commit/8aa1388bacef6ea40fbc5a394d7b191d42a867a6

mattslane (approved):

PR #490, opened 2022-10-21
BREAKING CHANGE: allowing usage of oidc instead of aws keys if specified

RostamKal:
upversioning to circleci/aws-ecr@8.1.3 and allowing usage of oidc instead of aws keys/secrets if specified
This version uses oidc for the relevant jobs if the oidc parameter is set to true. Otherwise it will look for AWS keys as before by default
The newer aws-ecr version uses a newer docker version + buildx which doesn't seem to work with the deprecated github container repo url (docker.pkg.github.com). It fails to pull images with 401's despite a login using the same url. 
So i've changed the login to use the new url (ghcr.io). This is the breaking part of the change and requires any github container repo urls in dockerfiles to be amended to ghcr.io

### **NEW CONTEXT VARIABLES USED**
CIRCLE_IAM_ROLE_ARN (optional)
AWS_ACCOUNT_ID (already in jaws contexts)

### **DEPRECATED CONTEXT VARIABLES**
ACCOUNT_URL

### DIFF
--- a/jaws-journey-deploy/changelog.md
+++ b/jaws-journey-deploy/changelog.md
@@ -1,0 +1,18 @@
+# 2.0.0
+
+* aws-ecr orb changed from version `7.2.0` to `8.1.3`
+* ovotech/aws-configure-credentials-oidc@1.0.1 added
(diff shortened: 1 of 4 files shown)


stephen-harris (approved):

robturnerovo (approved):
awesome, lgtm

PR #493, opened 2022-11-03
Masc 410 argo sync request

RostamKal:
A sync request can be sent before polling the status as before


argocd/sync_request copy.sh
```diff
@@ -0,0 +1,14 @@
+
```
stephen-harris, line 1: Should thise file be committed? Looks like a test script :)

argocd/scripts/sync_and_wait.py
```diff
@@ -0,0 +1,133 @@
...
+
+    except ValueError as e:
+        print(f'ERROR: Decoding JSON has failed: {e}')
+        return False
+        
+    except Exception as e:
+        print(f'ERROR: Request failed: {e}')
+        return False
+
+def is_cluster_insync(endpoint, token, application, target_revision):
```
stephen-harris, line 39: I think we've duplicated some python code. Could we combine in into one python file, and then just invoke either wait or sync based on command line arguments?

argocd/scripts/output.json
```diff
@@ -0,0 +1,377 @@
+{
+    "metadata": {
```
stephen-harris, line 2: Should this have been commited? Same for output-jaws-api.json?

stephen-harris (changes requested):

argocd/scripts/sync_and_wait.py
RostamKal: Good point, given this ended up being a minimal change in the end, I've now put the sync into the same job as before. No new job or scripts

stephen-harris (approved):

PR #506, opened 2022-12-19
Updates rac-gcp-deploy

quartin:
* [Adds target-file](https://github.com/ovotech/circleci-orbs/commit/5d318fc48bca91326c1e8cdf048202f56baa62bd) Since c&e projects don't have a standard Dockerfile in root, snyk won't be able to auto-detect it and provide base image remediation advice.

* [Exposes monitor-on-build parameter](https://github.com/ovotech/circleci-orbs/commit/f9d8fabb5b6b826fe61f6cc13087c921e881174c)

### DIFF
--- a/rac-gcp-deploy/orb.yml
+++ b/rac-gcp-deploy/orb.yml
@@ -167,6 +167,10 @@ commands:
       snyk-project-name:
         description: "Name of the target Snyk project"
         type: string
+      monitor-on-build:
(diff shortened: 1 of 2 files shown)


jensraaby (approved):
I can't test this but it looks safe 👍  (there is a way to take an orb from a git branch and use it in a repo, but I can't remember how)

PR #550, opened 2023-02-21
kap-kmi-deploy: add region param to job

clarencedglee:
Adds ability to specify which region (e.g. "eu1", "ap1") to update gitops on.

In Workflow we want to be able to release to oea-test and ovo-uat at the same time, have a manual approval step, and then progress the build to oea-sandbox and ovo-prod. We'd need to be able to specify the region to deploy to ovo-uat without affecting oea-sandbox (which is uat in git-ops).

### DIFF
--- a/kap-kmi-deploy/README.md
+++ b/kap-kmi-deploy/README.md
@@ -55,6 +55,7 @@ workflows:
           name: kmi-deploy-uat
           environment: uat
           push-image: true
+          kmi-region: eu1
       - kmi-deploy/deploy:
           name: kmi-deploy-prod
           environment: prod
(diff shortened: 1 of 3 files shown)


kelveden (changes requested):
The new parameter is being added but not used in the `update-gitops` command.

clarencedglee:
Isn't it being used here?
https://github.com/ovotech/circleci-orbs/pull/550/files#diff-1a0bf0ce605edc51359d38d1a2eda65fda156109cb661a52c1e1801391201351R175

And then picked up here?
https://github.com/ovotech/circleci-orbs/pull/550/files#diff-1a0bf0ce605edc51359d38d1a2eda65fda156109cb661a52c1e1801391201351R50

Maybe I'm reading it wrong

kelveden (approved):

PR #572, opened 2023-05-25
JAWS orb: bump Snyk orb version

thatfulvioguy:
Bumping again to use the update where curl will retry downloading the Snyk CLI in more cases

### DIFF
--- a/jaws-journey-deploy/changelog.md
+++ b/jaws-journey-deploy/changelog.md
@@ -1,3 +1,6 @@
+# 3.0.2
+* Bump Snyk orb version to get retries for the Snyk CLI download step in more cases
+
 # 3.0.1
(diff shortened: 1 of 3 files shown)


thatfulvioguy:
Putting back into draft for now until a new orb version is released which uses a recent enough version of curl to actually support the additional retries

thatfulvioguy:
Ready for review now. The problem with the orb was that since `snyk/install` is just a command and not a job, the executor is determined by the enclosing job (i.e. us, with the old curl version being due to some jobs remaining on Ubuntu 20.04 machine images :grimacing:). Upstream now have a fallback to not retrying all errors if the curl version is too old, which means builds will not break and we'll later get the benefits for free when we get round to bumping the machine image versions.

PR #618, opened 2024-08-16
Prevent disclosure of registry credentials in logs in the Terraform V2 orb

chris-brindley:
This PR disables xtrace when testing for presence of and handling the environment variable `$TF_REGISTRY_TOKEN`, which is a sensitive value.


darren-thomas (approved):
LGTM

